PR #14, opened 2021-07-31
Use remote and revision in manifest

electrikjesus:


anbox.xml
```diff
@@ -5,30 +5,33 @@
   <remove-project name="platform/system/libhwbinder" />
   <remove-project name="platform/external/libdrm" />
   <remove-project name="platform/external/mesa3d" />
+<!--
   <remove-project name="platform/hardware/intel/common/libva" />
```
erfanoabdi, line 9: this commented section is a test or some leftover?

electrikjesus: I forgot about that one... we need to not remove it for Bliss, but remove it for Lineage. So we need to get creative there. I have a card up my sleeve for that though

electrikjesus: I will plan on moving the manifest to a vendor method that we can generate the remove-project parts from dynamically

electrikjesus: This change has been wrapped into vendor_waydroid now. https://github.com/waydroid/android_vendor_waydroid/commit/c502d2bbfab0881a64e43c5210f3e213caa44c8d

Closing here

electrikjesus:
This change has been wrapped into vendor_waydroid now. https://github.com/waydroid/android_vendor_waydroid/commit/c502d2bbfab0881a64e43c5210f3e213caa44c8d

Closing here

PR #15, opened 2021-08-01
Some shell related clean-ups (untested)

sickcodes:
I think the big sudo EOF should be separated as it's hard to maintain, let me know what you think


erfanoabdi:
Thanks for your contribution really appreciate it but this scripts all are going to be removed soon

sickcodes:
It's still a good script for automation later when you want to do automated builds :)

PR #18, opened 2021-08-07
xorg check/warning

sinanmohd:


sickcodes:
@sinanmohd I don't think this is useful because if I have a wayland window open in X11 it would prevent it from running, and also the output for me is `tty`:

```console
[user@hostname ~]$ loginctl show-session "$XDG_SESSION_ID" -p Type --value
tty
```

sinanmohd:
@sickcodes yeah you're right i made this before that, are you using startx?, i tested this in a couple of just werks distros, anyway isn't this script  gonna be replaced by erfanoabdi/waydroid soon

sickcodes:
> @sickcodes yeah you're right i made this before that, are you using startx?, i tested this in a couple of just werks distros, anyway isn't this script gonna be replaced by erfanoabdi/waydroid soon

This works for desktop: https://github.com/sickcodes/droid-native

Issue #84, opened 2021-09-24
Some waydroid commands fails when run as normal user just because of logging


sigmaSd:
This line actually requires root https://github.com/waydroid/waydroid/blob/bullseye/tools/__init__.py#L44 because it appends to the log file.
This makes commands like waydroid session start or show-full-ui fail when run as normal user just because of logging. (commenting that like makes them work normally)

sigmaSd:
The issue boils down to `/var/lib/waydroid/waydroid.log` becomes inaccessible to normal user somehow.

steps to reproduce:
- `sudo systemctl stop waydroid-container`
- `rm /var/lib/waydroid/waydroid.log`
- `sudo systemctl start waydroid-container`
- `stat /var/lib/waydroid/waydroid.log` -> 0o666
- Run `waydroid session start` for a couple of seconds then ctrlc
-  `stat /var/lib/waydroid/waydroid.log` -> 0o700

sigmaSd:
Current workaround 
`sudo chown urusername /var/lib/waydroid/waydroid.log`

JordanViknar:
I can confirm I also encountered this problem. It started happening after doing "waydroid upgrade" and updating waydroid using the AUR. I suspect the problem is related to either the newest waydroid update on the AUR or the "waydroid upgrade" command.

sigmaSd:
Fixed in https://github.com/waydroid/waydroid/commit/9cec7cac9481750a4955c5bbc6c68a218d4635db

DaniyalAhmadSE:
I'm facing the same issue, now i can only run waydroid commands as `sudo`

natesworks:
this was not fixed (workaround works)

natesworks:
```
natesworks@pi:~ $ waydroid log
Run 'waydroid log' for details.
```

works as root

PR #85, opened 2021-09-24
Log to stdout if the user is not root

sigmaSd:
https://github.com/waydroid/waydroid/issues/84

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -13,8 +13,11 @@ from .helpers import logging as tools_logging
 
 
 def main():
+    def isRoot():
+         return os.geteuid() == 0
+
     def actionNeedRoot(action):
-        if os.geteuid() != 0:
+        if not isRoot():
             raise RuntimeError(
                 "Action \"{}\" needs root access".format(action))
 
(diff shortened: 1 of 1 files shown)


sigmaSd:
Or I can change the log path to /tmp/tools.log I'm not sure which is better.

sigmaSd:
I just noticed if I remove `/var/lib/waydroid/waydroid.log` and `waydroid init` `/var/lib/waydroid/waydroid.log` will be created with `0666` permission (which makes this pr unneeded).

So maybe somehow I had the wrong permission on this file, any idea?

sigmaSd:
If waydroid used to setup different permissions, I think that creating the log file should now check for the correct permission and if its incorrect, recreate the file.

sigmaSd:
I suspect https://github.com/waydroid/waydroid/blob/bullseye/tools/config/save.py is changing the permission.

So initially when the log is  created its accessible to the user, then after saving it becomes inaccessible making commands like show-full-ui fail.

sigmaSd:
The problem boils down to https://github.com/waydroid/waydroid/blob/bullseye/tools/config/save.py changing /var/lib/waydroid/waydroid.log permission.

I think that what should be fixed instead.

PR #86, opened 2021-09-24
Umount at session end

sigmaSd:


erfanoabdi (changes requested):
LGTM, just update commit msg to "container: Umount on stop"

PR #95, opened 2021-09-25
container: expose /dev/video0

sigmaSd:
The camera only worked after adding this in my case.


sigmaSd:
Multiple graphic cards can be present, if this pr is accepted maybe I should add entries for all of them (`find /dev -name "video*"`)

erfanoabdi:
https://github.com/waydroid/waydroid/blob/bullseye/tools/helpers/lxc.py#L61

PR #102, opened 2021-09-26
Add binder devices to gbinder conf

sigmaSd:


erfanoabdi:
no it's wrong and breaks halium devices

sigmaSd:
@erfanoabdi without this arch users currently have to create /etc/gbinder.conf with api 29 for waydroid to work.

Is that a correct approach?

erfanoabdi:
yeah creating /etc/gbinder.conf is correct
i think it's better to put the conf file on separate package
gbinder itself isn't part of waydroid anyway

PR #112, opened 2021-09-28
container: workaround weird log permission bug

sigmaSd:


erfanoabdi:
how about `os.chmod(args.log, 0o666)`

sigmaSd:
done

sigmaSd:
I'm not 100% sure but I think the permission still changes by some othe paths in the code.

I think the safest thing is to put the chmodding here https://github.com/waydroid/waydroid/blob/bullseye/tools/helpers/lxc.py#L247

erfanoabdi:
merged

PR #116, opened 2021-09-29
user_manager.py: use one variable for apps dir

### LINKED ISSUE #109: .desktop file creation fails if ~/.local/share/applications does not exist
I'm not sure if XDG specifies who is responsible to create this directory (i.e. OS or the one who creates the first .desktop file).

plata:
fixes #109

**NOTE:** not tested

### DIFF
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -9,6 +9,8 @@ from tools.interfaces import IPlatform
 
 
 def start(args, unlocked_cb=None):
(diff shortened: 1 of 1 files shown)


plata:
It might be a good idea to use [PyXDG](https://pyxdg.readthedocs.io/en/latest/basedirectory.html#xdg.BaseDirectory.xdg_data_home) instead of the hard coded path.

erfanoabdi:
thanks
Will switch to XDG later

PR #124, opened 2021-10-01
session: don't panic when removing old userdata if its not empty

sigmaSd:
https://stackoverflow.com/questions/303200/how-do-i-remove-delete-a-folder-that-is-not-empty


erfanoabdi:
well thing is it should never be "not empty"
we moved contents before

sigmaSd:
I guess I somehow triggered this code which means that somehow I had waydroid under ~/.local/share and ~/home at the same time.
Well I'll leave it, if other people trigger this path as well, you may want to check this again

erfanoabdi:
hmm well if that happened to you
lets not risk and merge..

PR #127, opened 2021-10-02
tools: improve permission handling

sigmaSd:


erfanoabdi:
maybe someone has wayland on root, in that case seesion must run as root

erfanoabdi:
also app actions doesn't need root

sigmaSd:
> maybe someone has wayland on root, in that case seesion must run as root
Yeah, too bad this would have caught a lot of misuse tough

> also app actions doesn't need root
In my testing app install/remove fails without root (but launch/list are ok)

sigmaSd:
@erfanoabdi I figured why launch/remove needs sudo

Its because of ~/.local/share/waydroid/data

Initially its created with 777, but after Android writes to it, it becomes 771, which make it un-writable.

Is this the intended behavior?

erfanoabdi:
oof permissions, no that's not intended and 771 is weird
i think umask is getting changed at some point, no idea

sigmaSd:
@erfanoabdi What if I change the runtime error to a warning? I think its true that people can have wayland as root but I don't think its the common case (and its a warning anyway). And I'll remove the subactions path for now, those I guess need an actual fix for the  data permission

PR #132, opened 2021-10-03
init: check hardware requirement

sigmaSd:
The archwiki  link can be replaced by a link to waydroid docs when it gets a section about software rendering.


erfanoabdi:
i think it's better to move checkRequirement to https://github.com/waydroid/waydroid/blob/bullseye/tools/actions/initializer.py#L84

Danct12:
We can also have it to check if the kernel supports binder.

sigmaSd:
@erfanoabdi Do you think its a good idea to add kernel modules detection? code: https://github.com/waydroid/waydroid/compare/bullseye...sigmaSd:more_checks?expand=1

tools/actions/initializer.py
```diff
@@ -80,8 +81,41 @@ def setup_config(args):
     cfg["waydroid"]["hwbinder"] = args.HWBINDER_DRIVER
     tools.config.save(args, cfg)
 
+def checkRequirement():
+    """ Check the requirements for running waydroid"""
+    def cpu():
+        with open("/proc/cpuinfo", "r") as f:
+            data = f.read()
+            if not ("ssse3" in data and "sse4_1" in data and "sse4_2" in data):
```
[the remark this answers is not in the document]
sigmaSd: Does other architectures have different required instructions? And if so is there a list somewhere that have this information?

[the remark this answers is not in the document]
Johnnynator: Yeah, that one should be okay.
So check for nothing additonal on `aarch64`/`arm64`
For 32bit arm check Features: `vfpv3d16` and that CPU architecture field is at least 7 (thumb-2 is always existant on armv7 and newer)
and for 32bit x86 check for `ssse3` and `sse3`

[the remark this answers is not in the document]
sigmaSd: @Johnnynator is this correct https://github.com/waydroid/waydroid/pull/132/commits/a3f8622e76aa129c0712658cce7b9b84b8bad2ec ?

tools/actions/initializer.py
```diff
@@ -80,8 +80,31 @@ def setup_config(args):
...
     tools.config.save(args, cfg)
 
+def checkRequirement():
+    """ Check hardware requirement for running waydroid"""
+    def cpu():
+        requirement = {
+                "x86": ["ssse3","sse3"],
+                "x86_64": ["ssse3","sse3","sse4_1","sse4_2"],
+                "arm64": [],
+                "arm": ["vfpv3d16"],
```
erfanoabdi, line 90: i'm not sure about this, here's an arm phone can run waydroid: https://paste.ubuntu.com/p/MKWT5pNnz6/

sigmaSd: I guess VFPv3d16 is just called VFPv3

Johnnynator: Vfpv3 != VFPv3d16

I think `/proc/cpuinfo` just doesn't report everything on erfanoadbi device.

We would have to use `getauxval` and  check if the `HWCAP_VFPv3D16` bit is set.

sigmaSd: Indeed that's what  recommended here https://developer.android.com/ndk/guides/cpu-features

Unfortunately that's annoying to call in python

PR #142, opened 2021-10-08
arch: detect armv8l as 32bit arm

Johnnynator:


Johnnynator:
`uname -m` returns `aarch64` on `64bit` and `armv8l` if run as 32bit.
I'm not sure if it is guaranteed that a 32bit armv8l host can run a 64bit container. (It usually should be the case)

PR #184, opened 2021-10-25
Count tries of get_service and let user prematurely interrupt

devenblake:
I'm not familiar with waydroid's code so please excuse me if this is a sucky contribution.

I'm having an issue where the `Failed to get service {}, trying again...` message keeps re-appearing, but when I CTRL+C I get a long traceback so it makes it hard to see the original error message. It'd be nice to be able to cancel that retry loop because a thousand tries is a long time to wait for it to fail.

It'd also be nice to see how many times it's tried.

I have not tested this code.

### DIFF
--- a/tools/interfaces/IPlatform.py
+++ b/tools/interfaces/IPlatform.py
@@ -283,7 +283,7 @@ def get_service(args):
     while(not remote):
         if tries > 0:
             logging.warning(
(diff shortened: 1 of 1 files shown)


electrikjesus:
I cannot replicate the Failed to get service {}, trying again...  error.  But this does add the KeyboardInterrupt, which is helpful

JamiKettunen:
@aleasto Do you think this is ready for merge?

Issue #72, opened 2021-09-23
`waydroid init` should check requirements


nagayev:
Hello. I'm running Debian 11 KDE based and I forget I'm using X11 not Wayland.
But when I run `waydroid init` it just started downloading lineage os, without throwing any warnings like:
[WARNING] You are using X11, not Wayland. More info: waydro.id/errors/x11
Also waydroid can checks python3 and lxc.

erfanoabdi:
waydroid init can download images regardless of x11 or wayland
but actually starting session will throw warnings 
lets say for example you can init on x11 and switch to wayland and start waydroid session

PR #235, opened 2021-11-22
Show error to non-wayland users

pktiuk:
This commit addresses issue: https://github.com/waydroid/waydroid/issues/72

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -50,6 +50,7 @@ def main():
 
         if not os.environ.get("WAYLAND_DISPLAY"):
             logging.error("Wayland session not detected. Only Wayland session manager is supported.")
+            logging.error("If you're unable to use wayland you can use the weston compositor instead.")
             return 1
 
         # Initialize or require config
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -48,6 +48,10 @@ def main():
 
         tools_logging.init(args)
 
+        if os.environ.get("XDG_SESSION_TYPE")!="wayland":
+            logging.error("Wayland session not detected. Only Wayland session manager is supported.")
+            return 1
+
         # Initialize or require config
         if args.action == "init":
             actionNeedRoot(args.action)
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -48,7 +48,7 @@ def main():
 
         tools_logging.init(args)
 
-        if os.environ.get("XDG_SESSION_TYPE")!="wayland":
+        if not os.environ.get("WAYLAND_DISPLAY"):
             logging.error("Wayland session not detected. Only Wayland session manager is supported.")
             return 1
 


tools/__init__.py
```diff
@@ -48,6 +48,10 @@ def actionNeedRoot(action):
 
         tools_logging.init(args)
 
+        if os.environ.get("XDG_SESSION_TYPE")!="wayland":
```
mkg20001, line 51: this likely will break using for example weston compositor

mkg20001 (changes requested):

tools/__init__.py
pktiuk: Maybe?
```python
 if os.environ.get("XDG_SESSION_TYPE")=="x11":
```

mkg20001: The problem is that weston does not set XDG_SESSION_TYPE

image

mkg20001: There seems to be another way to check if a wayland server is running and can be connected (how the apps check it - since gedit for ex starts in weston when it's there, even when launched outside)

It could be checked that way instead

hcsubser: There is an environment variable called WAYLAND_DISPLAY if it is set that means that a wayland compositor is running, that should be used to check instead of XDG_SESSION_TYPE

pktiuk: @mkg20001  
Could you check it?

mkg20001: yes, that one works

so it should be

```suggestion
        if not os.environ.get("WAYLAND_DISPLAY"):
```

image

tools/__init__.py
```diff
@@ -48,6 +48,10 @@ def actionNeedRoot(action):
...
+            logging.error("Wayland session not detected. Only Wayland session manager is supported.")
```
mkg20001, line 52:
```suggestion
            logging.error("Wayland session not detected. Only Wayland session manager is supported.")
            logging.error("If you're unable to use wayland you can use the weston compositor instead.")
```

pktiuk:
I think it can be merged now :D

electrikjesus:
I will add it to the test list for this week then

electrikjesus (approved):
This works as expected in testing. Approved on my end

erfanoabdi:
this breaks ubuntu touch

mkg20001:
@erfanoabdi how? An exception could be made or some more checks could be done

pktiuk:
> @erfanoabdi how? An exception could be made or some more checks could be done

It is easier to just dismiss contribution instead of giving valuable feedback how to fix it.

Fuseteam:
because ubuntu-touch does not set `$XDG_SESSION_TYPE` to wayland, it sets it to `mir` while still supporting wayland apps

PR #241, opened 2021-11-26
scripts: make sure misc subfolder exists

FlorianFranzen:
Ran into some issue with this on NixOS. This should avoid any problems in the future. Only creates the misc folder, as `varlib` should exists already.


electrikjesus (approved):
Approved on my end. Works as expected

PR #242, opened 2021-11-27
lxc.py: Detect usage of VM and set correct gralloc value

Herrie82:
When using Qemu/VirtualBox, gralloc is set to "gbm" while it should be "default". Add py_vmdetect to detect if we're running in a virtual machine and set gralloc accordingly.

Signed-off-by: Herman van Hazendonk <github.com@herrie.org>


Herrie82:
https://github.com/waydroid/waydroid/pull/243 offers are more flexible and generic approach for overrides

PR #243, opened 2021-11-27
Add an optional [properties] in waydroid.cfg

Tofee:
On some platforms, like a VM, we don't have the "getprop" android
utility available. It could be, also, that the default values
for some properties don't match the target device.

With this PR, one can add a [properties] section in waydroid.cfg,
that will act as an override for the values put by waydroid.

Example, on our LuneOS x86 VM:
[properties]
ro.hardware.gralloc=default
ro.hardware.egl=mesa

A minimal waydroid.cfg could even be shipped with waydroid, and
the rest of the configuration will be filled with "waydroid init".

Signed-off-by: Christophe Chapuis <chris.chapuis@gmail.com>

### DIFF
--- a/tools/config/load.py
+++ b/tools/config/load.py
@@ -26,6 +26,10 @@ def load(args):
                           " default value from config: {}".format(cfg['waydroid'][key]))
             del cfg["waydroid"][key]
 
+    if "properties" not in cfg:
+        cfg["properties"] = {}
+    # no default values for property override
+
     return cfg
 
 def load_session():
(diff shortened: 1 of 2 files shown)


Tofee:
My PR is a bit early, I still have some tests to run (though the idea will stay the same)

tools/helpers/props.py
```diff
@@ -13,7 +13,11 @@ def host_get(args, prop):
         command = ["getprop", prop]
         return subprocess.run(command, stdout=subprocess.PIPE).stdout.decode('utf-8').strip()
     else:
-        return ""
+        cfg = tools.config.load(args)
+        if cfg["properties"][prop] is not None:
+            return cfg["properties"][prop]
+        else
```
Herrie82, line 19: Missing a : here

Tofee:
This PR has now been correctly tested, and works well on our side

electrikjesus (approved):
This will work for overriding props well. Approved on my end

PR #246, opened 2021-12-03
Prevent session start as root

mkg20001:
the user might accidently end up doing session start from a sudo shell
prevent it, otherwise we end up with our files being owned by root and the application working not so well

### DIFF
--- a/tools/actions/session_manager.py
+++ b/tools/actions/session_manager.py
@@ -15,6 +15,15 @@ def start(args, unlocked_cb=None):
         stop(args)
         sys.exit(0)
 
+    home = os.getenv("HOME")
+    # the user might accidently end up doing session start from a sudo shell
+    if home != "/root" and os.getuid() == 0:
+        # prevent it, otherwise we end up with our files being owned by root and the application working not so well
+        logging.error("You're running session start as root but home is not /root")
+        logging.error("session start should only be run as a regular user")
+        logging.error("If you insist on running it as root, ensure HOME is set to /root")
+        return 1
+
     xdg_session = os.getenv("XDG_SESSION_TYPE")
     if xdg_session != "wayland":
         logging.warning('XDG Session is not "wayland"')


electrikjesus:
There are changes needed for this to work as intended. Please look at my corrections here and resubmit https://github.com/electrikjesus/anbox-halium/commit/5054fe29bd973698fd8f871711a937b0314467ef

electrikjesus:
Actually, ignore my correction, I misunderstood what this was checking on.

mkg20001:
So it's just the `and` instead of `&&`? Crazy language this python

Fuseteam:
what is blocking this PR?

tools/actions/session_manager.py
```diff
@@ -15,6 +15,15 @@ def signal_handler(sig, frame):
         stop(args)
         sys.exit(0)
 
+    home = os.getenv("HOME")
+    # the user might accidently end up doing session start from a sudo shell
+    if home != "/root" and os.getuid() == 0:
```
septatrix, line 20: What if the root user has a different home directory? Might be better to check if the owner of `$HOME` is equal to the current user id

septatrix (changes requested):

tools/actions/session_manager.py
mkg20001: This is basically never the case, but I get your point, this might also cover other possible edge cases.

kj7rrv: Why isn't just `os.getuid() == 0` enough?

septatrix: Because this is intended to prevent accidentally running this as user when `HOME` is still the directory of an unprivileged user. Some distributions configure `sudo` to preserve this variable so starting this with sudo would result in writing files as root in the users home directory. Normally running this as root however should still be allowed which is why only `UID == 0` is too strict

aleasto:
Waydroid stores the android's data in home which inevitaly includes files owned by root. I'm not sure this helps.

Fuseteam:
isn't that why @septatrix suggested checking the owner of $HOME matches the current user?

JamiKettunen:
What about similarly preventing accidental runs of [`show-full-ui`](https://github.com/waydroid/waydroid/blob/1eba546/tools/actions/app_manager.py#L106) or [`first-launch`](https://github.com/waydroid/waydroid/blob/1eba546/tools/actions/initializer.py#L199), or should those be handled by separate PR(s) due to being out-of-scope?

Fuseteam:
i think a separate PR would be better depending on how complicate it would make this one

tools/actions/session_manager.py
JamiKettunen: Could leave a `FIXME` comment above here that `root` could have a different `$HOME` as well and that something like `getent passwd root | cut -d: -f6` should be used to get it instead of hardcoding it as `/root`.

JamiKettunen: @mkg20001 would you mind adding the `FIXME` as per my previous comment from a few months back? alternatively you could also implement getting the root user's `$HOME` properly too instead of hardcoding it as `/root`

JamiKettunen:
@aleasto Aside from [my comment](https://github.com/waydroid/waydroid/pull/246#discussion_r1000611095) still being unresolved, do you think this is ready for merge?

aleasto:
No, this is fundamentally wrong.

`sudo env | grep HOME` gives `/root`, making this check useless

JamiKettunen:
Ah nvm I see the issue now, this only "fixes" the issue on distributions where `sudo -E` is the default (examples being???), anyway this indeed needs to be reworked or else it prevents using Waydroid when signed in as root user and running a Wayland session etc.

PR #264, opened 2022-01-01
tweak tools.helpers.http.download()

aditya24raj:
while downloading, progress bar shows file sizes in bytes. it prints large numbers and makes it harder to understand the info.
this commit-
1.) shows file sizes in megabytes and an approximate download speed along with it. hoping it would make the download information easier to understand.

2.) adds "daemon=True", in the thread which starts progress(). it kills the progress thread if download thread is killed by ctrl+c. previously one had to press ctrl+c twice, one to kill downloading and one to kill progress bar thread.


electrikjesus (approved):
This works as expected. Approved on my end

Issue #77, opened 2021-09-23
Get rid of dnsmasq


atmouse-:
Dnsmasq need passthrough the apparmor rules or selinux which most of linux dist included by default.

We can use [DHCPServer] provided by systemd-networkd instead like this
```
[Match]
Name=waydroid0

[Network]
Address=192.168.250.1/24
DHCPServer=true
IPForward=yes

[DHCPServer]
PoolOffset=100
PoolSize=20
EmitDNS=yes
#DNS=8.8.8.8
```

systemd itself is the default init system of most dist ubuntu debian arch also Halium

Nidrop:
Disagree. I want to use waydroid in my Void linux (runit). Also i noticed some Alpine (openrc) users here in issues

Quackdoc:
This wound up falling under the radar, this should possible now by modifying `/var/lib/waydroid/lxc/waydroid/config` changing the interface to waydroid0 to something else, this will allow you to setup your own networks instead of relying on waydroid-net to do so. 

if you wanted to do this, feel free to make a brief write up and I would be happy to make a PR to add it to the docs.

PR #267, opened 2022-01-03
make waydroid-net more flexible

Quackdoc:
this should fix anbox interface conflict, as well as let users define their own bridge for instance if they want direct accsess to the host network. Ideally, it should get the interface adapter name from lxc config, as that would let users use their own pre-existing adapters without extra configuration (IE. virbr0 that kvm users commonly make), but for now this quick fix should not only alleviate some issues, but also for instance remove the hard dependency on dnsmasq addressing #77 so long as they are comfortable defining their own interface.

it's only been lightly tested. but it's working for me.


Quackdoc:
hmmm... seems like waydroid0 doesn't destroy itself, rather just disables itself when turned off. which is causing an issue when waydroid service gets restarted as it doesn't re-init, even when the interface goes up.

this is in general pretty bad behavior. but I'm not sure what solution would be best to remedy it.

Quackdoc:
should be ready for testing this will exit out of the waydroid script sorry for taking so long on this, one of the things that endlessly got put off.

data/scripts/waydroid-net.sh
```diff
@@ -2,14 +2,22 @@
...
 varrun="/run/waydroid-lxc"
 varlib="/var/lib"
+varvnic=$(awk '$1 == "lxc.net.0.link" {print $3}' /var/lib/waydroid/lxc/waydroid/config)
+
+if [ "$varvnic" != "waydroid0" ]; then
+        echo "varnic is $varvnic"	
+	exit 0
+else 
+	echo "varvnic is waydroid0"
+fi
```
aleasto, line 12: I don't think you meant to commit these.
Not the `exit 0` at least.

Quackdoc: exit 0 is used to exit script early. and the echo statements or so if someone messes around with it and needs help, we don't need to spend ages debugging it. 

this makes it so the users can create their own interface that waydroid script doesn't control at all. this removes a lot of the networking hard-deps like dnsmasq for say if someone wants to use dhcpserver from systemd-networkd. it can at least give package maintainers some flexibility for if a distro chooses not to ship dnsmasq for whatever reason.

the other usecase for this is defining network interfaces for things like ipvtap or macvtap or simply using pre-existing network bridges like libvirts virbr0.

and of course the last use makes it easy to just setup a no network waydroid.

Quackdoc: unless I messed something up, by default this shouldn't have an impact on normal usage

aleasto: Ack

Fuseteam:
oh this might help on ubuntu touch and on my laptop running a windows kvm :thinking: how do i test this?

Quackdoc:
> oh this might help on ubuntu touch and on my laptop running a windows kvm thinking how do i test this?

simply download the file and replace the file here `/usr/lib/waydroid/data/scripts/waydroid-net.sh`

aleasto:
@Quackdoc mergeable?

Quackdoc:
LGTM

Fuseteam:
oh my

PR #288, opened 2022-01-27
lxc: Optionally bind-mount NFC config

fredldotme:
- Devices with libnfc-nci might have their NFC HAL crash when
  the config doesn't exist or contains invalid data. Fix it.

Fixes NFC HAL crashing on UT on the Pixel 3a.


fredldotme:
I think this is ready too and shouldn't interfere with devices/setups which don't have libnfc-nci.conf.

PR #302, opened 2022-02-09
Depend on versioned LXC

smurfix:
Older LXC versions don't know about the `-P` option.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -13,7 +13,7 @@ Package: waydroid
 Architecture: all
 Depends: ${misc:Depends},
          ${python3:Depends},
-         lxc,
+         lxc (>= 1:4.0.10),
          python3-gbinder,
          python3-gi
 Description: Android™ application support


erfanoabdi:
-P is supported on older than 4.0.10 versions
at least 2.x supports it, which is used on ubuntu touch

JamiKettunen:
Yeah no this is just wrong because on Ubuntu Touch 16.04 (LXC v2.0.x) everything seems to be working just fine, what `lxc-info --version` was the problematic host having? Which exact command caused the error message and what was the output as well?

From what I can tell the `-P` / `--lxcpath` option was introduced [way back in LXC v0.9.0.rc1](https://github.com/lxc/lxc/commit/67e571de63a8e465dc8f1b17e16744a1d3fb552c) 10 years ago, I don't think we should even support LXC v2 anymore really since Ubuntu Touch 20.04 exists and is soon having an initial stable release.

PR #320, opened 2022-03-10
tools: Fix device names with spaces in initializer

fredldotme:
Don't try to fetch from an URL with a space inside.


erfanoabdi:
Wondering whats ro.product.device on jingpad that has space!?
isn't it illegal to put space in device codename?

fredldotme:
That's with Jing's default vendor image, don't know what they screwed up here.
Might check and overlay things in the meantime, but you are right, it actually sounds like illegal behavior.

erfanoabdi:
i don’t mind merging this change, was just curious lol

Fuseteam:
what currently blocking this PR?

erfanoabdi:
sorry, i forgot

Issue #46, opened 2021-09-13
[Feature request] List of apps not to add launchers for


Djhg2000:
On Mobian the list of launchers in Phosh grows quite a lot when installing Waydroid. On the other hand we probably still want the launchers for apps installed by the user.

This is a feature request to have an option in the `/var/lib/waydroid/waydroid.cfg` config file where we can list some apps not to add launchers for.

As an example I'm thinking of something like this:
```
[launcher]
no_launcher = com.android.documentsui,com.android.contacts,org.lineageos.recorder
```

If I'm reading the code right the list would then get compared somewhere around [line 24 in waydroid/tools/services/user_manager.py](https://github.com/waydroid/waydroid/blob/d2f25933af662bfa168b9d510444b95ab1fa976d/tools/services/user_manager.py#L24), but I'm not entirely sure about that.

An alternative approach could be to use the `packageName` as a key and the value to specify options for that particular launcher, in case some launchers need other special options passes to the session manager in the future:
```
[launchers]
com.android.documentsui = no_launcher
com.android.contacts = grayscale_icon
org.lineageos.recorder = stop_session_on_exit,grayscale_icon
```

The options other than `no_launcher` are just examples to serve as placeholders. This alternate approach would be more flexible but it's probably more tedious to implement.

philmmanjaro:
Maybe this is a way to hide Waydroid Apps on **Phosh**: https://gitlab.manjaro.org/manjaro-arm/packages/community/waydroid/-/commit/720ea28f30f65cc4128e2e3458bedbd45ecac18b

erfanoabdi:
that kills all app icons tho 🤔

Djhg2000:
Indeed, a better workaround for now is [what I documented on the Mobian Wiki](https://wiki.mobian-project.org/doku.php?id=waydroid#optional-tweaks); `truncate -s 0 ~/.local/share/applications/waydroid.[package_name].desktop`

philmmanjaro:
On **Phosh** we don't need the icons in the drawer. Totally fine within the Waydroid launcher.

TimMagee:
As a phosh user who really hopes to see Android apps integrate well with my system, I really want to see my Android apps show up in the Phosh app drawer.

Would is be possible to simply remove these apps from our Android ROM? 

[...]

PR #339, opened 2022-03-28
Create .desktop files for system apps with "NoDisplay=true"

plata:
see #46

This approach blacklists the default system apps which should be sufficient for the vast majority of use cases. If required, it can be extended in future PRs to have a configurable blacklist.

### DIFF
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -34,11 +34,13 @@ def start(args, unlocked_cb=None):
         packageName = appInfo["packageName"]
 
         desktop_file_path = args.apps_dir + "/waydroid." + packageName + ".desktop"
-        if not os.path.exists(desktop_file_path) and not packageName in system_apps:
+        if not os.path.exists(desktop_file_path):
             lines = ["[Desktop Entry]", "Type=Application"]
             lines.append("Name=" + appInfo["name"])
             lines.append("Exec=waydroid app launch " + packageName)
             lines.append("Icon=" + args.waydroid_data + "/icons/" + packageName + ".png")
+            if packageName in system_apps:
+                lines.append("NoDisplay=true")
             desktop_file = open(desktop_file_path, "w")
             for line in lines:
                 desktop_file.write(line + "\n")
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -9,6 +9,19 @@ from tools.interfaces import IPlatform
 
 
 def start(args, unlocked_cb=None):
+    system_apps = [
+        "com.android.calculator2",
+        "com.android.documentsui",
+        "com.android.settings",
+        "org.lineageos.jelly",
+        "com.android.contacts",
+        "com.android.email",
+        "org.lineageos.eleven",
+        "org.lineageos.recorder",
+        "com.android.deskclock",
+        "com.android.gallery3d",
+        "org.lineageos.etar"
+    ]
 
     def makeDesktopFile(appInfo):
         showApp = False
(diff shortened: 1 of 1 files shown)


Fuseteam:
some of these apps seem useless for most users, i wonder if can just rebuild the image without these apps 🤔

dos1:
It may be a better idea to use `NoDisplay=true` instead of not creating .desktop files at all.

plata:
@dos1 I've changed the code. It should work but I cannot test it currently.

dos1:
On Android 11 image with microG I got also these created right after first boot, it may be a good idea to add them to the list too:
```
com.android.camera2
com.android.inputmethod.latin
com.google.android.gms
```

plata:
Done.

Doomsdayrs:
This is a good idea, since a lot just clutter the menu

Quackdoc:
I'm not really happy with this, since many users will prefer to have these instead. maybe we could punt this into a config, and add a script to the waydroid cli to add it to existing images?

Fuseteam:
All these? I doubt many users want to see the android calculator xD
i do see many wanting settings tho hmm

Doomsdayrs:
I can state that I do not like typing in "settings" and opening up android settings.

Fuseteam:
That can be optimized in another way, there are plenty of users that utilize setting to configure android settings ;)

Quackdoc:
> All these? I doubt many users want to see the android calculator xD i do see many wanting settings tho hmm

calculator is actually explicitly one of the things I often use. which just further proves my point. I don't think any of these should be completely kicked off.

I think using a flag to disable the generation of files and a way for waydroid cli to set the display flag would be optimal.

Fuseteam:
> calculator is actually explicitly one of the things I often use. which just further proves my point. I don't think any of these should be completely kicked off.

I am under the assumption that most users run waydroid on a system that already has a preinstalled calculator i suppose xd (now they have 2 all of a sudden)

Quackdoc:
> > calculator is actually explicitly one of the things I often use. which just further proves my point. I don't think any of these should be completely kicked off.
> 
> I am under the assumption that most users run waydroid on a system that already has a preinstalled calculator i suppose xd (now they have 2 all of a sudden)

I do too, I just prefer using android's which is where the main issue comes in. we shouldn't be forcing something either way on people who dont want it when we a configuration would be suitable for doing it

Fuseteam:
As long as that configuration file thingie doesn't require root i guess. Otherwise it be useless for a certain call of users (aka not in /etc please)

Personally i don't like all the duplicate apps that comes with waydroid, i'd rather the image be build without all the duplicate apps. I would like think it would be easy for users to install them if one needs/wants them. But hey that's just me, I'm just a minority

plata:
> But hey that's just me, I'm just a minority

I don't think so.

The default use case for Waydroid is (in my opinion) to run apps which are only available for Android. Somewhat like Wine to that regard. Therefore, I guess it's safe to assume that the average user wants to see the explicitly installed apps and not what comes merely as part of the runtime.

From what I'm concerned, most likely more people are currently annoyed by the desktop entries than would miss them.

Fuseteam:
yeah same, I think users rather install applications they use. applications that can't be installed from an app store should be visible tho, like android settings. the user should know what is installed on their pc

JamiKettunen:
To get this moving forward perhaps it would be good to at least introduce a `persist.waydroid.hide_system_apps` prop which users can set if they prefer to not see these?

Perhaps having even the list of system apps itself configurable would make sense, though as seen with e.g. `persist.waydroid.fake_touch` on https://docs.waydro.id/usage/waydroid-prop-options there is a 91 character limit, so a prop couldn't exactly be used for that :/

siddhpant:
Bump, this is needed, I don't want to open Waydroid when I quickly type calc and press enter.

PR #341, opened 2022-03-28
Install icons to hicolor

rmnscnce:
This PR installs the 512x512 Waydroid icon into the standard Freedesktop.org `hicolor` fallback icon theme. This enables the use of icon names in the `.desktop` file, as exhibited in a commit inside this PR.

In addition, the location of the icon has been moved into a more organized place. Changes necessary to keep the app from breaking has been done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ WAYDROID_DIR := $(PREFIX)/lib/waydroid
 BIN_DIR := $(PREFIX)/bin
 APPS_DIR := $(PREFIX)/share/applications
 METAINFO_DIR := $(PREFIX)/share/metainfo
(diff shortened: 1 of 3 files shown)


erfanoabdi:
looks good to me on desktop but need to be tested on ubuntu touch too

rmnscnce:
Should be applicable to Ubuntu Touch as well, unless if the shell in UT doesn't respect Freedesktop.org specifications.

JamiKettunen:
@aleasto Would you be interested in getting this in after testing on UT?

JamiKettunen:
oh right conflicts need to be resolved first, @rmnscnce mind rebasing?

Issue #238, opened 2021-11-24
Terminal apps don't work with lxc 4.0.11.


arun54321:
Distro: Arch linux
Waydroid Image: android 11

FlorianFranzen:
Termux crashes with `java.lang.RuntimeException: Cannot open /dev/ptmx` for me too.

Etaash-mathamsetty:
also having this issue

chqoot-vgxu:
The issue is caused by lxc not punting devpts on /dev/pts. I still haven't figured out why it doesn't get mounted though.

A workaround would be to manually mount it with `mount -t devpts -o defaults,mode=644,ptmxmode=666 devpts /dev/pts`

Etaash-mathamsetty:
> The issue is caused by lxc not mounting devpts on /dev/pts. I still haven't figured out why it doesn't get mounted though.
> 
> A workaround would be to manually mount it with `mount -t devpts -o defaults,mode=644,ptmxmode=666 devpts /dev/pts`

I looked at the code and am clueless as well

Etaash-mathamsetty:
how can i run it on waydroid startup?

chqoot-vgxu:
You shouldn't, that's why this is a workaround and not a fix. Unfortunately I didn't have time to investigate further for a real fix

Quackdoc:
this should be fixed now

Etaash-mathamsetty:
> this should be fixed now

still broken for me on waydroid 1.3.0

chqoot-vgxu:
> still broken for me on waydroid 1.3.0 error: cannot open /dev/ptmx
Updating waydroid doesn't necessarily update lxc config. Check if `lxc.pty.max = 10` is present in `/var/lib/waydroid/lxc/waydroid/config`, if not add it manually

Etaash-mathamsetty:
> > still broken for me on waydroid 1.3.0 error: cannot open /dev/ptmx
> 
> Updating waydroid doesn't necessarily update lxc config. Check if `lxc.pty.max = 10` is present in `/var/lib/waydroid/lxc/waydroid/config`, if not add it manually

works fine now!

Issue #145, opened 2021-10-10
Issues opening waydroid shell


devenblake:
When executing `echo "ip route add default via 192.168.250.1" | sudo waydroid shell` as advised in the [Waydroid postmarketOS Wiki page](https://wiki.postmarketos.org/wiki/Waydroid) I got the following error:
```
lxc-attach: waydroid: terminal.c: lxc_terminal_create_native: 927 Not a directory - Failed to open terminal multiplexer device
```
This error seems to occur with any attempt to run `waydroid shell` as root, in ssh and in the device's terminal emulator. Here's the error message when running `waydroid shell` as root without a command or input stream:
```
lxc-attach: waydroid: terminal.c: lxc_terminal_create_native: 927 Not a directory - Failed to open terminal multiplexer device
/system/bin/sh: No controlling tty: open /dev/tty: No such file or directory
/system/bin/sh: warning: won't have full job control
```
I don't see anything relevant in `waydroid log`. I also tried `chown`ing the log file back to root, running `waydroid shell` as root, and then running `waydroid log as root`, but I didn't get any entries in the log other than the fact that I was accessing it.

This does not occur when running `waydroid shell` as user but it won't let me do anything as user:
```
[08:52:03] ERROR: Action "shell" needs root access
[08:52:03] See also: <https://github.com/waydroid>
Run 'waydroid log' for details
```

I'm using a Pinephone on postmarketOS+Phosh edge, updated/upgraded yesterday.

spoorun:
Are you using Arch or Manjaro @devenblake ?
If so, check the work-around here: 

I.e. instead of 
sudo wayland shell
use:
sudo lxc-attach -P /var/lib/waydroid/lxc/ -n waydroid --clear-env

i.e. 

sudo lxc-attach -P /var/lib/waydroid/lxc/ -n waydroid --clear-env
[sudo] password for simpx: 
/bin/sh: No controlling tty: open /dev/tty: No such file or directory
/bin/sh: warning: won't have full job control
:/ # ip route add default via 192.168.250.1
:/ # exit

I still didn't get internet to work, but the setting took without error...

spoorun:
Got it working now - 
sudo nmcli device delete anbox0
sudo nmcli con add type bridge ifname waydroid0 -- connection.id waydroid-net ipv4.method shared ipv4.addresses 192.168.250.1/24
restart waydroid session

hakimzulkufli:
[...]

PR #349, opened 2022-04-06
Allows waydroid to create PTY

Quackdoc:
doing this will allows LXC to create their own PTY instead of mounting host, this should be more secure then mounting /dev/pty 

addresses 

#145 #238

### DIFF
--- a/data/configs/config_2
+++ b/data/configs/config_2
@@ -19,7 +19,7 @@ lxc.net.0.hwaddr = 00:16:3e:f9:d3:03
 lxc.net.0.mtu = 1500
 
 lxc.console.path = none
-xc.pty.max = 10
+lxc.pty.max = 10
 
 lxc.include = /var/lib/waydroid/lxc/waydroid/config_nodes
 
--- a/data/configs/config_2
+++ b/data/configs/config_2
@@ -19,6 +19,7 @@ lxc.net.0.hwaddr = 00:16:3e:f9:d3:03
 lxc.net.0.mtu = 1500
 
 lxc.console.path = none
+xc.pty.max = 10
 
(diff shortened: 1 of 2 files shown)


Quackdoc:
10 is just an arbitrary number, I don't see any user exceeding this.

data/configs/config_2
```diff
@@ -19,6 +19,7 @@ lxc.net.0.hwaddr = 00:16:3e:f9:d3:03
 lxc.net.0.mtu = 1500
 
 lxc.console.path = none
+xc.pty.max = 10
```
aleasto, line 22: typo, but also this would just be duplicate.

Quackdoc: oops, but doesn't this need to be reflected in both files?

aleasto: Sorry, that's right. I thought these were fragments then merged into a single config file.

Etaash-mathamsetty:
can this get merged asap please, would like this fixed

JamiKettunen:
This broke launching `waydroid-container` on Ubuntu Touch 16.04 (soon to be retired in favor of 20.04 where there is no problem):
```
root@ubuntu-phablet:~# lxc-info --version
2.0.11
root@ubuntu-phablet:~# waydroid container start
lxc-info: confile.c: parse_line: 2452 Unknown configuration key "lxc.pty.max"
lxc-info: parse.c: lxc_file_for_each_line: 57 Failed to parse config: lxc.pty.max = 10

Failed to load config for waydroid
Failure to retrieve information on /var/lib/waydroid/lxc:waydroid
[20:08:54] WayDroid container is
```

PR #372, opened 2022-05-06
Wait for init

aleasto:
In the interest of simplifying the installation of waydroid, make it possible to start the container manager before running init.
The intended usecase is to have waydroid-container automatically enabled and started on install. This was previously impossible because the service would immediatly stop because waydroid was not initialized.

### DIFF
--- a/debian/waydroid-container.service
+++ b/debian/waydroid-container.service
@@ -2,7 +2,7 @@
 Description=Waydroid Container
 
 [Service]
-ExecStart=/usr/bin/waydroid container start
+ExecStart=/usr/bin/waydroid -w container start
 ExecStop=/usr/bin/waydroid container stop
 ExecStopPost=/usr/bin/waydroid session stop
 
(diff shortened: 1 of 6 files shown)


Fuseteam:
if we can make waydroid initialize itself it wasn't initialized before starting the container, i see no reason to hide it behind a flag. why not make that the default behavior?

aleasto:
It's not initializing itself. Initialization requires user intervention to choose the system and vendor channels, as well as the system type (gapps etc...)

It's just sleeping until initialization.

Fuseteam:
oh cool but that could still be the default behavior rather than a simple error message xd

Fuseteam:
not sure what changed but lemme re-iterate:
i don't think we need a flag to wait for init, i think it should wait for init by default instead of giving an error message xd

aleasto:
@erfanoabdi what do you think?
Can you make the deb installer start and enable the service cause i don't know how to do that

aleasto:
Originally I implemented this through fifo pipes to avoid polling the disk, but i realize everything else in waydroid comunicates by polling the disk.

Do you want to introduce this `ipc.py` to maybe later use it elsewhere and save some disk access and power, or it's overkill and i should just `while not actions.initializer.is_initialized(args):` ?

PR #374, opened 2022-05-11
Mount TUN device for use by Android VPNs.

KenMacD:
Providing access to the hosts `/dev/net/tun` is required by at
least some VPN clients, for example F5 Access.


erfanoabdi:
Should be like this: `make_entry("/dev/net/tun", "dev/tun")`

KenMacD:
@erfanoabdi Thank you. Fixed and rebased.

PR #387, opened 2022-05-20
add checklist to issue template

Quackdoc:
This should allow anyone helping with an issue assume that the person creating the issue has checked for duplicates, has read the docs, and is infact using wayland. any issues created missing one of these checks should have a good reason for doing so, or just close the ticket


JamiKettunen:
This is good but I think even better would be to convert the whole thing to [an issue form](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) as I mentioned in #478 to make it even more user-friendly.

PR #405, opened 2022-06-12
Set aidl version based on the android version

aleasto:


tools/actions/container_manager.py
```diff
@@ -81,11 +82,51 @@ def chmod(path, mode):
...
+            mnt = "/tmp/waydroid-" + str(uuid.uuid1())
+            helpers.mount.mount(args, cfg["waydroid"]["images_path"] + "/system.img", mnt)
+            with open(mnt + "/system/build.prop") as build_prop:
+                for line in build_prop:
+                    line = line.strip()
+                    if len(line) == 0 or line[0] == "#":
+                        continue
+                    k,v = line.partition("=")[::2]
+                    if k == "ro.build.version.sdk":
+                        android_api = v;
```
JamiKettunen, line 98: `android_api = int(v)` so we can do integer logic with the API level? ideally you should also handle the potentially generated `ValueError` if doing this though as well just in case.

tools/actions/container_manager.py
```diff
@@ -81,11 +82,51 @@ def chmod(path, mode):
...
+            if android_api == "":
+                raise Exception()
+        except:
+            logging.error("Failed to parse android version from system.img")
+        finally:
+            helpers.mount.umount_all(args, mnt);
+
+        binder_protocol = "aidl"
+        sm_protocol = "aidl"
+        if android_api == "28" or android_api == "29":
```
JamiKettunen, line 109: Imho ideally this whole block would look something like taking my previous integer logic suggestion into account:
```py
if android_api < 28:
    binder_protocol = "aidl"
    sm_protocol = "aidl"
elif android_api < 30:
    binder_protocol = "aidl2"
    sm_protocol = "aidl2"
elif android_api == 30:
    binder_protocol = "aidl3"
    sm_protocol = "aidl3"
else:
    binder_protocol = "aidl3"
    sm_protocol = "aidl4"
```

tools/actions/container_manager.py
```diff
@@ -81,11 +82,43 @@ def chmod(path, mode):
...
+                    "ro.build.version.sdk"))
+        except:
+            logging.error("Failed to parse android version from system.img")
+        finally:
+            helpers.mount.umount_all(args, mnt);
+
+        if android_api < 28:
+            binder_protocol = "aidl"
+            sm_protocol =     "aidl"
+        if android_api < 30:
```
cniw, line 101: shouldn't we use `elif` too at this line?

cniw (changes requested):

PR #418, opened 2022-06-18
app_manager: Attempt to wait for window to spawn

fredldotme:
Wait up to 10 seconds for the window to spawn before exiting.

### DIFF
--- a/tools/actions/app_manager.py
+++ b/tools/actions/app_manager.py
@@ -61,6 +61,14 @@ def launch(args):
                     2, "policy_control", "immersive.full=*")
         else:
             logging.error("Failed to access IPlatform service")
+        attempts = 0
+        max_attempts = 10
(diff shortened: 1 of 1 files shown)


aleasto:
Ideally this would be a callback received through binder, without polling getprop. I'm not sure how feasible it is, since there would be multiple processes waiting for a message about their own app, so there has to be some sorting...

Also, what about the first app that launches with session stopped: that process stays alive until the session ends. Doesn't that break your model?

aleasto:
Untitled Diagram drawio
This is how I would structure it using #662
Each app_manager registers a dbus object with path equal to the app package name. The session manager is the single recipient of binder calls from the waydroid compositor and wakes up the correct app_manager.

fredldotme:
That looks like a rather interesting solution indeed. Yeah that would fit the description of what we need.

aleasto:
Currently blocked by https://github.com/mer-hybris/libgbinder/issues/111

PR #419, opened 2022-06-18
Add session watchdog

aleasto:
Needs https://github.com/waydroid/android_vendor_waydroid/pull/9 to be reliable, otherwise sometimes on very fast systems the session callback is never triggered and the watchdog would bark.

### DIFF
--- a/tools/actions/session_manager.py
+++ b/tools/actions/session_manager.py
@@ -7,8 +7,22 @@ import signal
 import sys
 import shutil
 import tools.config
+from threading import Timer
 from tools import services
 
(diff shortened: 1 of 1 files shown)


JamiKettunen:
I suppose this is ready now as the linked PR is merged?

aleasto:
> I suppose this is ready now as the linked PR is merged?

Not really, this requires that all users upgrade images, so it'll be stuck for a while.

Also 5 min watchdog kinda defeats the purpose, imho. Will see if i can do better.

JamiKettunen:
Considering the change was in for 17.1 images this could be pursued again

PR #426, opened 2022-06-26
Downgrade x86_64 to x86 if no SSE4.2

aleasto:


aleasto:
also lays the ground for detecting aarch64-only architectures

Issue #412, opened 2022-06-15
market:// protocol not opening Google play store


apy7hst:
When I use the market:// protocol to open the google play store in ubuntu touch, it does not open, neither does it open on ubuntu 20.04. Please fix this minor issue.

PR #428, opened 2022-06-28
Add Android Market Protocol Handler

cniw:
When clicking market:// protocol on Linux distro host, then open it on Waydroid.

Related: https://github.com/waydroid/waydroid/issues/412


aleasto:
I've implemented that.
https://github.com/waydroid/waydroid/pull/433
https://github.com/waydroid/waydroid/issues/412

`waydroid app intent android.intent.action.VIEW %u`

PR #439, opened 2022-07-06
Add first-launch command

aleasto:
Add a graphical init to be run as root, which we can use in
first-launch through pkexec


aleasto:
This is how it looks
Screenshot_20220714_213436

PR #440, opened 2022-07-06
Add Makefile buildsystem

JamiKettunen:
This aims to solve the packaging mess currently in the wild with each distribution doing essentially duplicate work manually copying the needed files/dirs to the correct locations; also avoids new changes being missed in some distro packaging when everyone just `make install`s the same things.


aleasto:
Couple of requests from fedora:
* USE_SYSTEMD is not configurable from outside
* Add something to switch waydroid-net to nftables, like `sed -i '/LXC_USE_NFT=/s/false/true/' $(WAYDROID_DIR)/data/scripts/waydroid-net.sh`

JamiKettunen:
> USE_SYSTEMD is not configurable from outside

@aleasto how precicely? `make install USE_SYSTEMD=0` works for me, or is it something else missing? like being able to set `USE_SYSTEMD=0` in env and just `make_install`?

aleasto:
Yea. same for `PREFIX` maybe.
Should be enough to switch to `?=` iirc

JamiKettunen:
Yep `?=` will do it, what about all the others (`*_DIR` vars)? I mean it shouldn't really hurt to make them changeable via env should it?

aleasto:
Ehh i don't know. There's some hardcoded paths in waydroid. Changing PREFIX probably already breaks stuff

JamiKettunen:
Ah yes, there's some hardcoded stuff currently:
- `systemd/waydroid-container.service` hardcodes `/usr/bin/waydroid`
- `data/{waydroid.market,Waydroid}.desktop` both harcode icon as `/usr/lib/waydroid/data/AppIcon.png` (different icon sizes when along with an svg? :p)
- `tools/config/__init__.py` harcodes `/usr/share/waydroid-extra`
(found with a simple grep for `/usr`)

Should anything be done about this? What about disallowing change of `PREFIX` for now?

aleasto:
Eh, keep it as is.

aleasto:
fyi i've added a new file, `./polkit/*`

JamiKettunen:
Tested the `debian/*` changes enough to see it produces similar enough contents (`dpkg -c waydroid_*_all.deb `) to the old packaging, think this is ready for review and proper testing now.

aleasto (approved):

aleasto:
One thing I want to make sure is that waydroid-container is still automatically enabled and started on install

PR #454, opened 2022-07-18
Support upgrading major android versions

aleasto:
When restarting the lxc container our binder services will receive a
binder_presence event and will try to re-register on the service manager.
If the service manager was created with aidl2 and the new android version
uses aidl3 this will fail with:
  Failed to add service waydroidusermonitor: -2147483647

When that happens, try to restart the service by getting a new service
manager that can handle the new aidl protocol version.

### DIFF
--- a/tools/actions/container_manager.py
+++ b/tools/actions/container_manager.py
@@ -82,33 +82,6 @@ def start(args):
         for path in perm_list:
             chmod(path, mode)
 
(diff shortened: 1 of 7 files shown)


apy7hst:
I think LineageOS is a very good alternative of android out there, it is better to maintain waydroid as lineageOS itself.

PR #455, opened 2022-07-19
Add first-launch command [v2]

aleasto:


aleasto:
So this time the initialization starts in the container process removing the need to prompt the user for authorization via polkit.
The progress output is piped to the gui client that draws it on screen.

There is no `waydroid init --gui` anymore but that's a fair price to pay.

Still TODO: switch to a ui toolkit that uses wayland.

CC: @JamiKettunen

JamiKettunen:
Wouldn't we want a new dependency in `debian/control` for `gtk+3.0`? Don't know if `python3-gi` already pulls this on Debian/Ubuntu systems already

aleasto:
Yes. Do you know the correct package name? `gtk3`?

JamiKettunen:
It doesn't look like either `gtk3` or `gtk+3.0` exist on ubuntu focal, but it looks like [`libgtk-3-0`](https://packages.debian.org/search?keywords=libgtk-3-0) might fit the bill

JamiKettunen:
Ah I found out what it really needs: `/usr/lib/girepository-1.0/Gtk-3.0.typelib` is the file on my Void system that it looks for with `gi.require_version("Gtk", "3.0")`, and on Debian systems it is provided by [`gir1.2-gtk-3.0`](https://packages.debian.org/search?keywords=gir1.2-gtk-3.0)

PR #459, opened 2022-07-23
Add appstream metadata

aleasto:


aleasto:
@electrikjesus can I assume the screenshots to be CC0-1.0?

data/org.waydro.id.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.waydro.id</id>
+
+  <name>Waydroid</name>
+  <summary>Run android apps natively alongisde desktop programs</summary>
```
cniw, line 6: typo should be "alongside" and "Android" with capital 'A'

data/org.waydro.id.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.waydro.id</id>
```
cniw, line 3: Since Waydroid's domain is waydro.id, why it's not reversed become id.waydro or if you prefer use org just org.waydroid without second dot

data/org.waydro.id.metainfo.xml
```diff
@@ -0,0 +1,36 @@
...
+    </p>
+  </description>
+
+  <launchable type="desktop-id">Waydroid.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://waydro.id/assets/images/hero/main_landing.jpeg</image>
+    </screenshot>
+    <screenshot>
+      <image>https://waydro.id/assets/images/hero/Group%201.png</image>
```
cniw, line 30: Better use underscore instead of space and use lowercase, so it's the same as any other URL

data/org.waydro.id.metainfo.xml
```diff
@@ -0,0 +1,36 @@
...
+  <supports>
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>touch</control>
+    <control>tablet</control>
+  </supports>
+
+  <description>
+    <p>
+      Waydroid is a a container-based approach to boot a full Android system on a regular GNU/Linux system. The Android system inside the container has direct access to needed hardware through LXC and the binder interface.
```
cniw, line 20: The 'a' is double in the third word

data/org.waydro.id.metainfo.xml
apy7hst: You can ask about this to the members of waydroid

data/org.waydro.id.metainfo.xml
apy7hst: It should be "linux desktop programs" instead of "desktop programs"

data/org.waydro.id.metainfo.xml
aleasto: I'm reusing existing urls.
It doesn't matter.

data/id.waydro.waydroid.metainfo.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>id.waydro.waydroid</id>
+
+  <name>Waydroid</name>
+  <summary>Run Android apps natively alongisde desktop programs</summary>
```
cniw, line 6: still typo should be "alongside"

PR #490, opened 2022-08-18
Add mainline vulkan support

aleasto:
Attempts to set the correct ro.hardware.vulkan value based on the
loaded drm driver name

### DIFF
--- a/tools/helpers/gpu.py
+++ b/tools/helpers/gpu.py
@@ -4,10 +4,26 @@ import tools.helpers.props
 
 unsupported = ["nvidia"]
 
+def getKernelDriver(args, dev):
+    return tools.helpers.props.file_get(args, "/sys/class/drm/{}/device/uevent".format(dev), "DRIVER")
+
 def getDriNode(args):
     for node in glob.glob("/dev/dri/renderD*"):
         dev = os.path.basename(node)
-        driver = tools.helpers.props.file_get(args, "/sys/class/drm/{}/device/uevent".format(dev), "DRIVER")
-        if driver not in unsupported:
+        if getKernelDriver(args, dev) not in unsupported:
             return node
+    return ""
+
+def getVulkanDriver(args, dev):
+    mapping = {
+        "i915": "intel",
+        "amdgpu": "radeon",
+        "radeon": "radeon",
+        "panfrost": "panfrost",
+        "msm": "freedreno",
+        "vc4": "broadcom",
+    }
+    kernel_driver = getKernelDriver(args, dev)
+    if kernel_driver in mapping:
+        return mapping[kernel_driver]
     return ""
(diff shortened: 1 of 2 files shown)


Quackdoc:
do we have images that support this for testing?

aleasto:
> do we have images that support this for testing?

No but they can be compiled from current git

aleasto:
I should add freedreno, broadcom and panfrost too.
And it should probably fallback to lavapipe by default

Quackdoc:
is it worth falling back to lavapipe by default? software vk is probably fine on most modern x86_64 machines, but I worry about arm devices specifically, as there are plenty of low end arm devices floating around with linux, could this wind up causing even more issues then VK just not working on said devices? 

if so maybe default for x86 devices but leave arm to not use lavapipe, at least by default.

aleasto:
What issues are you thinking about? Not much uses vulkan

Since we are declaring that the device [supports vulkan](https://github.com/waydroid/android_device_waydroid_waydroid/blob/lineage-17.1/device.mk#L155-L157) i think it is actually better to make sure we always provide an implementation.

Quackdoc:
on significantly low powered devices, running lavapipe, or other software rendering stuff for that matter can cause severe issues due to the cpu usage, and I know some people are making it run on absurdly ancient hardware. this could present problems in android itself rendering incredibly slowly, or if unlucky, even causing issues for the host OS. for such users falling back to lavapipe could be worse then having no support for vulkan. this could cause issues for people running games with vulkan support. which there are a few on the gplay store. 

though I'm not sure a good way to go about dealing with this one to be honest. would not adding the driver be that bad? iirc vulkanloader should still report that no devices work anyways and fail elegantly. this would be enough to prevent apps from auto detecting vulkan support and using it over opengl assuming it doesnt cause issues. 

Also I might be mistaken since I'm a wee tired seems to be checking for kernel driver for deciding what vulkan driver to load? there are plenty of radeon cards and intel cards that would fall through the gaps with this. not declaring lvp and not supporting vulkan either.

Ideally we could query `VkPhysicalDeviceProperties` using vulkan but that would require using vulkan stuff in the script iirc granted im not familiar with python so that might not actually be too bad? don't know enough to comment. 
another possible solution would be to query vulkaninfo too, but then we add a dependency vulkaninfo...

hm... I don't think this is a good solution, but I don't know of any better ones myself. if this is what waydroid ends up going with, I think an entry in the docs detailing both how to disable lavapipe, or enable it would be necessary to accompany the merge.

aleasto:
> seems to be checking for kernel driver for deciding what vulkan driver to load? there are plenty of radeon cards and intel cards that would fall through the gaps with this

Elaborate?

aleasto:
https://www.phoronix.com/news/Intel-HASVK-Old-Vulkan-Gen7-8 This will be very annoying though.

aleasto:
Will deal with intel when the driver split lands in mesa-stable

Issue #491, opened 2022-08-19
Container will not start on system startup


ACR-Jeff:
**Describe the bug**
Waydroid Container will not start on system startup. I have to manually start or restart the container to get waydroid to work. I have tried everything I could think of.
My temporary solution is creating a script, With the contends of 
`sudo systemctl restart waydroid-container
sudo systemctl restart waydroid-container`

This is only a temporary solution as it isn't an actual solution, It allows waydroid to work, Yet each time I have to enter my password.
Any suggestions would be great. 

**Screenshots**

**General information (please complete the following information):**
 - Waydroid tools Version [e.g. 1.1.0]
 - Waydroid Images Version [e.g. VANILLA 20211021]

**Desktop (please complete the following information):**
 - OS: Debian Bullseye 11
 - GPU: Intel Device 5906
 - Kernel version: 5.18.0-0.bpo-1-amd64
 - Host mesa version: [e.g. mesa 2.2]
 - Desktop: Gnome 1:3.38+3

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 3a]
 - OS: [e.g. Ubuntu touch OTA19]
 - Mainline or Halium [e.g. Halium 9]
 - Kernel version: [e.g. 4.14.x]

**Additional context**

**Logs (please upload as file)**
 - Prop file, located on `/var/lib/waydroid/waydroid_base.prop`
 - Output of `waydroid log`
 - *Full* (not partially) Output of `waydroid logcat` from the beginning of starting waydroid services

electrikjesus:
Take a close look here at home we are doing it on Waydroid-Linux https://github.com/waydroid/waydroid-linux_tools/tree/main/wd_startup

Issue #489, opened 2022-08-18
Document page installation instructions changed and are incorrect.


ACR-Jeff:
**Describe the bug**
I have installed waydroid a number of times the last few weeks. Today I went to the Documents page [](https://docs.waydro.id/usage/install-on-desktops) to remember the proper installation instructions and get the repo for Debian Bullseye install. Yet the instructions seemed to have changed within the last few hours. I am not sure if this is the only step missing or changed but I know for a fact `sudo waydroid init` has been removed (Maybe intentional). 
I followed the steps on the Documents page exactly and when getting to the part to run waydroid from the applications menu a window pops up with "Intialize Waydroid" Followed by the system ota and vendor ota, Yet when trying to download it just presents with "Waiting for waydroid container service..., The waydroid container service is not listening." So then I tried `sudo systemctl restart waydroid-container` and this solved that problem. This step maybe needed to be added to the instructions or code to prevent further issues. After `sudo systemctl restart waydroid-container` then going to the Waydroid application menu and running waydroid it works perfect. Thank you.

**Screenshots**

**General information (please complete the following information):**
 - Waydroid tools Version [e.g. 1.1.0]
 - Waydroid Images Version [e.g. VANILLA 20211021]

**Desktop (please complete the following information):**
 - OS: Debian Bullseye Gnome
 - GPU: [e.g. AMD Radeon™ RX 5700]
 - Kernel version: 5.10.0-17-amd64
 - Host mesa version: [e.g. mesa 2.2]
 - Desktop: [e.g. Gnome 40.x.x]

**Smartphone (please complete the following information):**
 - Device: [e.g. Pixel 3a]
 - OS: [e.g. Ubuntu touch OTA19]
 - Mainline or Halium [e.g. Halium 9]
 - Kernel version: [e.g. 4.14.x]

**Additional context**

**Logs (please upload as file)**
 - Prop file, located on `/var/lib/waydroid/waydroid_base.prop`
 - Output of `waydroid log`
 - *Full* (not partially) Output of `waydroid logcat` from the beginning of starting waydroid services

aleasto:
Did you have waydroid previously installed on this same debian installation?

ACR-Jeff:
> Did you have waydroid previously installed on this same Debian installation?

Nope this was a fresh install of Debian with only system updates.

PR #492, opened 2022-08-19
Fix debian service not autostarting

aleasto:
Fixes: https://github.com/waydroid/waydroid/issues/491
Fixes: https://github.com/waydroid/waydroid/issues/489


aleasto:
Verified that it was broken previously and that it works now

aleasto:
Fixes #491 #489

PR #502, opened 2022-08-29
Detect HIDL gralloc4

aleasto:


aleasto:
Requires:
* python >= 3.8
* gbinder-python >= 1.1.1

fallsback to not detecting gralloc4 if those requirements aren't met

aleasto:
I'm thinking if it might be enough to hide this behind `if args.vendor_type != "MAINLINE"` to trust that there will always be some hwbinder service registered so that even older python-gbinder won't crash

PR #503, opened 2022-08-30
Enable the "No New Privileges" restriction for the container

ChrysoliteAzalea:
Hello everyone! I propose enabling the "No New Privileges" restriction for the container. 

**Explanation**:

"No New Privileges" is a restrictive flag on a process that prevents changing the effective UID, GID or capability set during an **execve**. It also restricts MAC domain transition, and allows a process to enable seccomp and Landlock self-restriction without CAP_SYS_ADMIN.

**Possible benefits**:

* Preventing unwanted AppArmor domain transition
* Preventing SUID flag and other "privileged file" flag based security risks.

**Possible problems**:

* I don't know how NNP affects SELinux domain transitions. AppArmor has a guaranteed restrictive domain transition feature (profile stacking) that works with NNP enabled, but I don't know whether the same exists in SELinux.
* I don't know how NNP affects root in the container. If an application relies on a **su** binary with a SUID flag to get root, then NNP may break it. (UPD: Magisk root works with container-wide NNP enabled)
* I also don't know how NNP affects Google Play Services.

### DIFF
--- a/data/configs/config_2
+++ b/data/configs/config_2
@@ -7,6 +7,8 @@ lxc.autodev = 0
 # lxc.autodev.tmpfs.size = 25000000
 lxc.apparmor.profile = unconfined
 
+lxc.no_new_privs = 1
+
 lxc.init.cmd = /init
 
 lxc.mount.auto = cgroup:ro sys:ro proc


ChrysoliteAzalea:
Today, I've done some testing on rooted container (with Magisk), and, as far as I tested, the "No New Privileges" restriction doesn't affect the Magisk root. As far as I understand, it happens because the **su** binary in Magisk doesn't rely on the SUID bit or file capabilities -- it instead makes a request to the Magisk Manager app and if superuser access is granted, the app is relaunched as root, with full effective set of capabilities and without zygote-level seccomp.

However, it may break non-Magisk root solutions that rely on SUID or file capabilities. I think that we can mention it in the documentation.

PR #504, opened 2022-08-30
Restrict container capabilities

ChrysoliteAzalea:
Hello everyone! I propose restricting POSIX capabilities for root processes in the container. 

**Explanation**:

POSIX capabilities are a way for granular management of superuser privileges in GNU/Linux systems. They allow granting some superuser privileges to a non-root process and restricting superuser privileges for a root process.

**Possible benefits**:

Mitigating security risks for example, in case of vulnerabilities in root processes in the container (however, **this measure only is not enough** and has to be supplemented by other security enhancements).

**Possible problems**:

May affects some apps that need root. Will not affect ordinary apps since they already have their capability set cleared empty by the zygote.


ChrysoliteAzalea:
Today, I've tested it with Magisk root. Doesn't seem to affect anything.

PR #506, opened 2022-08-30
Add AppArmor profiles for the container.

ChrysoliteAzalea:
Hello everyone! I propose adding AppArmor profiles for the container.

**Explanation**:

AppArmor is a mandatory access control feature in the Linux kernel, an alternative to SELinux. It allows to restrict processes access to various resources, such as files, signals, tracing, POSIX capabilities, sockets, and other features.

**Possible benefits**:

* Mitigating security risks in case of vulnerabilities in root processes.
* Preventing container escape.

**Possible problems**:

* If a system lacks AppArmor, and exception will be raised and caught, notifying the user and continuing normal operation
* AppArmor profile will also confine root applications
* This hasn't been tested with Google Play Services
* May not work on old kernels that lack AppArmor profile stacking feature.

Because of this, all profiles are in **complain** mode. They may be later switched to **enforce** mode.


ChrysoliteAzalea:
AppArmor profiles from the original commit were breaking the Magisk root by preventing edits to /sbin. In order to fix that, I've added a read-write access rule for /sbin to the **lxc-waydroid** profile. This shouldn't affect Android apps since they are also confined with the **android_app** profile.

I also noticed that some process is trying to edit and/or lock some files in /system/framework/ directory every time when I grant root access to some application. It's denied on my system, and it doesn't break root, however, I don't know whether it's going to affect something.

ChrysoliteAzalea:
Force-pushed in order to resolve a conflict.

ChrysoliteAzalea:
These profiles were written for Android 10, and they seem to cause issues on Android 11 if they are in **enforce** mode. I'll try to figure out the issues and propose a fix.

ChrysoliteAzalea:
OK, I have modified the profiles to fit Android 11. The new policies work fine for me in **enforce** mode.

tools/helpers/lxc.py
```diff
@@ -155,6 +157,22 @@ def set_lxc_config(args):
     command = ["cp", "-fpr", seccomp_profile, lxc_path + "/waydroid.seccomp"]
     tools.helpers.run.user(args, command)
 
+    try:
+        command = ["cp", "-i", apparmor_profiles[0], apparmor_profile_dir + "lxc/lxc-waydroid"]
```
aleasto, line 161: Why `cp -i` in an automated script?
```
-i, --interactive
              prompt before overwrite
```
Also perhaps these could just become `shutil.copy`

aleasto (commented):
I'll have to test this on a system without apparmor, but for now leaving this review

tools/helpers/lxc.py
ChrysoliteAzalea: I've intended it to be a safety measure, that Waydroid installation wouldn't overwrite user's own security policy.

aleasto: We can't have that otherwise the policies won't be updated when we update waydroid

aleasto: Also errors to install the profiles should only be logged to `waydroid log` and not to stdout, at least on distros that don't use AppArmor.

Overall I'm leaning towards the idea that this should be implemented by distros downstream, like I implement SELinux for fedora downstream

ChrysoliteAzalea: > We can't have that otherwise the policies won't be updated when we update waydroid

Well, we can then just replace the ```-i``` with ```-f```, however, we **should** warn the user, because they can already have stricter security policy, and in that case, we can accidentally make it less strict, which doesn't seem a good idea to me.



> Overall I'm leaning towards the idea that this should be implemented by distros downstream, like I implement SELinux for fedora downstream.

The problem is that only distributions that package and ship Waydroid from their repositories (like Arch Linux, if we consider AUR to be their repository) can do that. Debian and some other distributions don't have Waydroid in their official repositories, and I doubt that they will ship AppArmor profiles for software they don't ship.

aleasto: Why couldn't waydroid.deb include files in `/etc/apparmor.d` ?

ChrysoliteAzalea: waydroid.deb can include these files (as well as ```apparmor_parser``` call from ```postinst``` in order to load them to the kernel), it just needs to be packaged that way.

aleasto: Yes, that's what i was suggesting by "Perhaps we can have a makefile target that installs the policies".

ChrysoliteAzalea: Alright, I'll try that. The only question is that the container should be in ```unconfined``` domain unless the policy is installed, and in ```lxc-waydroid``` if it is installed. Therefore, the make target should reconfigure the container. I think it could be done with ```sed```.

aleasto: Yes probably just sed

ChrysoliteAzalea: Done -- now all the profile management is done through the ```make apparmor``` target rather than LXC helper. The interactive overwrite was also replaced with force overwrite, since the policy is supposed to be managed by the Waydroid project (users can still use their own policy, they just need to choose different names for profiles).

Makefile
```diff
@@ -31,3 +31,14 @@ install:
 	if [ $(USE_NFTABLES) = 1 ]; then \
 		sed '/LXC_USE_NFT=/ s/false/true/' -i $(INSTALL_WAYDROID_DIR)/data/scripts/waydroid-net.sh; \
 	fi
+
+apparmor:
```
aleasto, line 35: `install_apparmor`

data/configs/adbd
```diff
@@ -0,0 +1,58 @@
+#include <tunables/global>
```
aleasto, line 1: Please make the filenames of apparmor configs recognisable (maybe `data/configs/adbd.apparmor` )

Makefile
```diff
@@ -31,3 +31,14 @@ install:
...
+	cp -f data/configs/adbd /etc/apparmor.d/adbd
+	apparmor_parser -r /etc/apparmor.d/adbd
+	cp -f data/configs/android_app /etc/apparmor.d/android_app
+	apparmor_parser -r /etc/apparmor.d/android_app
+	cp -f data/configs/lxc-waydroid /etc/apparmor.d/lxc/lxc-waydroid
+	apparmor_parser -r /etc/apparmor.d/lxc/lxc-waydroid
+	sed --sandbox -i "s/lxc.aa_profile = unconfined/lxc.aa_profile = lxc-waydroid/g;" /var/lib/waydroid/lxc/waydroid/config
+	sed --sandbox -i "s/lxc.apparmor.profile = unconfined/lxc.apparmor.profile = lxc-waydroid/g;" /var/lib/waydroid/lxc/waydroid/config
```
aleasto, line 43: `/lxc.aa_profile/ s/unconfined/lxc-waydroid/g`

aleasto, line 43: You cannot modify `/var/lib/waydroid` during the install as that is only created at runtime.
You need to modify the source config files in `$(DESTDIR)$(WAYDROID_DIR)/data/configs/`

Makefile
```diff
@@ -31,3 +31,14 @@ install:
 	if [ $(USE_NFTABLES) = 1 ]; then \
 		sed '/LXC_USE_NFT=/ s/false/true/' -i $(INSTALL_WAYDROID_DIR)/data/scripts/waydroid-net.sh; \
 	fi
+
+apparmor:
+	cp -f data/configs/adbd /etc/apparmor.d/adbd
```
aleasto, line 36: Define `APPARMOR_DIR := /etc/apparmor.d` (in the line below `SYSD_DIR`) and `INSTALL_APPARMOR_DIR := $(DESTDIR)$(APPARMOR_DIR)` (in the line below `INSTALL_SYSD_DIR`), then copy to it.

Basically copy files the same way they are copied in `install:`

aleasto (changes requested):
Thank you. I like this much better.
Left some comments to address

Makefile
ChrysoliteAzalea: Done.

data/configs/adbd
ChrysoliteAzalea: Done -- AppArmor configs are moved into an ```apparmor_profiles``` subdirectory.

Makefile
ChrysoliteAzalea: Done.

ChrysoliteAzalea: Done. However, I'd like to ask -- does that mean that ```install_apparmor``` has to be called *before* ```install```?

Makefile
ChrysoliteAzalea: Done.

Makefile
aleasto: The opposite, in fact.
`install` will install `$(DESTDIR)$(WAYDROID_DIR)/data/configs`, then `install_apparmor` can run sed on it

aleasto (approved):

aleasto (commented):
I think i misinterpreted `apparmor_parser`. I thought it was checking if the profile is valid.
Does it have to be run at runtime instead? Just once or at every boot?

aleasto:
Ok. Only copy the files in the makefile target, then packaging scripts will handle the enabling part:
https://wiki.debian.org/AppArmor/Contribute/FirstTimeProfileImport

ChrysoliteAzalea:
So, the Makefile should just copy the profiles? (I think that if a user installs a packaged Waydroid, the profiles are loaded by ```postinst```, but if a user installs it from GitHub, they should be loaded by ```make```).

I also wonder that, at which stage should ```lxc.aa_profile``` and ```lxc.apparmor.profile``` settings be changed from ```unconfined``` to ```lxc-waydroid```?

aleasto:
Made some changes, what do you think?
debhelper uses `apparmor_parser -r -T -W ` so i've switched to that "By using '-W -T' we ensure that any abstraction updates are also pulled in."

aleasto:
> I also wonder that, at which stage should ```lxc.aa_profile``` and ```lxc.apparmor.profile``` settings be changed from ```unconfined``` to ```lxc-waydroid```?

That is indeed a good question. Technically I don't like that the `install_apparmor` is modifying a file installed by `install` (concept of file ownership in packaging), but I don't personally see another way

aleasto:
PS: Shipping these profiles officially means there should be someone maintaining them... Hopefully you'll be around for future android version upgrades :)

ChrysoliteAzalea:
> Made some changes, what do you think? debhelper uses `apparmor_parser -r -T -W ` so i've switched to that "By using '-W -T' we ensure that any abstraction updates are also pulled in."

I think it's a good idea, because sometimes, abstractions in AppArmor profiles have to be changed (I've modified them on some of my systems).

> PS: Shipping these profiles officially means there should be someone maintaining them... Hopefully you'll be around for future android version upgrades :)

Sure, I also hope so!

aleasto:
Okk that should correctly handle all cases

aleasto:
PopOS 20.04 is being silly.
The systemd service for lxc is supposed to parse the lxc apparmor profiles on start, but it's failing for unrelated reasons:
```
AppArmor parser error for /etc/apparmor.d/lxc-containers in /etc/apparmor.d/tunables/home at line 16: syntax error, unexpected TOK_SET_VAR, expecting $end
```

So lxc-waydroid is not parsed and waydroid fails to start with:
```
lxc-start: waydroid: lsm/apparmor.c: apparmor_process_label_set_at: 1183 No such file or directory - Failed to write AppArmor profile "lxc-waydroid" to 53
lxc-start: waydroid: lsm/apparmor.c: apparmor_process_label_set: 1229 Invalid argument - Failed to change AppArmor profile to lxc-waydroid
```

PR #515, opened 2022-09-06
Rebuild waydroid.prop after live upgrade

aleasto:
On waydroid upgrade waydroid_base.prop might change so we should
rewrite the mountable waydroid.prop file as well.

Do this in mount_rootfs so that:
* Every caller already benefits from this (initializer, upgrader...)
* We can set props based on rootfs contents


erfanoabdi:
looks good

PR #531, opened 2022-09-18
Configure multiple preinstalled images paths

aleasto:
Precedence:
  /etc/waydroid-extra/images
  /usr/share/waydroid-extra/images

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -29,13 +29,16 @@ def setup_config(args):
     args.arch = helpers.arch.host()
     cfg["waydroid"]["arch"] = args.arch
 
-    preinstalled_images = tools.config.defaults["preinstalled_images_path"]
+    preinstalled_images_paths = tools.config.defaults["preinstalled_images_paths"]
     if not args.images_path:
-        if os.path.isdir(preinstalled_images):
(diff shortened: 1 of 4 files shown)


peat-psuwit:
Is the path configurable by end users?

aleasto:
> Is the path configurable by end users?

No. This was never meant to be an end-user feature but rather a way for distros to preinstall images. And my opinion is that it should remain the same.

However this became popular among users after the release of some third-party Android 11-based images.
I'm thus ok considering this an 'advanced user' or 'system administrator' feature, for which a predefined location makes sense (like config files for many unix programs) and is more than enough.
Within this scope, this change helps distributions with immutable or r-o `/usr` such as silverblue or ubports. For everybody else, I don't think this should be the primary way of using waydroid.

I've briefly considered changing the semantics of the already existing `--images_path` flag to achieve what you're asking, while maintaining the current meaning of _<where to download the OTA images into>_ but that is more complicated than I'm willing to invest time on, given my previous reasoning.

Finally, if you're working on some android images for waydroid and want to distribute them to end users, the correct way is to host an OTA server (a static json file over https works just fine) **and maintain your work through OTA updates**

PR #548, opened 2022-10-14
use host OS timezone inside of android

### LINKED ISSUE #542: Provide local time
**Describe the bug**
The UTC time is provided by default. Is there some way to propagate the timezone settings from the Linux host to the Android container?

**Screenshots**

**General information (please complete the following information):**
 - Waydroid tools Version: latest
 - Waydroid Images Version: latest GAPPS

**Smartphone (please complete the following information):**
 - Device: Pinephone
 - OS: Mobian (Debian Bookworm)
 - Mainline
 - Kernel version: 5.15
 - Host mesa version: 22.2.0
 - Desktop: Phosh

**Additional context**

**Logs (please upload as file)**

Quackdoc:
fixed #542

### DIFF
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -227,6 +227,11 @@ def make_base_props(args):
     if ccodec != "":
         props.append("debug.stagefright.ccodec=" + ccodec)
 
+    timezone = os.readlink("/etc/localtime")
+    if timezone != "":
+        timezone = timezone.replace("/usr/share/zoneinfo/","")
+        props.append("persist.sys.timezone=" + timezone)
+
     ext_library = tools.helpers.props.host_get(args, "ro.vendor.extension_library")
     if ext_library != "":
         ext_library = ext_library.replace("vendor/", "vendor_extra/")


tools/helpers/lxc.py
```diff
@@ -227,6 +227,11 @@ def find_hidl(intf):
     if ccodec != "":
         props.append("debug.stagefright.ccodec=" + ccodec)
 
+    timezone = os.readlink("/etc/localtime")
```
aleasto, line 230: Are there weird systems where `/etc/localtime` does not exist?
If not sure this better be wrapped in `try/except` or `os.path.exists`

Quackdoc: /etc/localtime is standard, though I suppose people who don't bother to set it up on their more DIY distros like arch might not have this, will update the PR later.

Quackdoc:
Im going to close this for now, since apparently there are a good number of distros in which this wont work. ill look for another method,

tools/helpers/lxc.py
darkdragon-001: You could read the contents of `/etc/timezone` instead. I would say this solves the problem for 99% of the users as all major distros support it even though there might be some old and obscure distros which don't support it.

Quackdoc: timezone isn't ubiquitous either sadly. ive encounter numerous installs with out them, even my personal arch install is lacking it

darkdragon-001: It's in the nature of Linux that since you can replace any package from distro to distro, you will never find something which works for all. I would still suggest to start with something which works for the majority of users. Later one can add other special cases as a fallback when the first one isn't available.

aleasto: /etc/timezone is not a thing in fedora either

Quackdoc: we could use (time.tzname) but that reports as for eastern instead of a location, (like Canada/Eastern for me for instance) `('EST', 'EDT')`, not sure if android will just accept that, and if so, how to format it

PR #550, opened 2022-10-15
Remove CAP_SYS_MODULE from the capability bounding set

ChrysoliteAzalea:
Hello everyone! I propose removing CAP_SYS_MODULE from the capability bounding set.

**Explanation**:

POSIX capabilities are a way for granular management of superuser privileges in GNU/Linux systems. They allow granting some superuser privileges to a non-root process and restricting superuser privileges for a root process. CAP_SYS_MODULE is a capability that allows a process to load and unload kernel modules. Removing it would prevent root processes from modifying the kernel.

**Possible benefits**:

Protect the kernel from compromised root processes and root apps.

**Possible problems**:

I don't see any -- kernel module loading is already blocked by a [container-wide seccomp filter](https://github.com/waydroid/waydroid/pull/505).

### DIFF
--- a/data/configs/config_base
+++ b/data/configs/config_base
@@ -5,7 +5,7 @@ lxc.arch = LXCARCH
 lxc.autodev = 0
 # lxc.autodev.tmpfs.size = 25000000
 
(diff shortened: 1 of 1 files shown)


Quackdoc:
I havent encountered an issue with this, myself.

ChrysoliteAzalea:
> I havent encountered an issue with this, myself. so no issues from me

Still, I seem to have added it by mistake. The Android container shouldn't load modules to the kernel (meanwhile, it's already blocked by a container-wide seccomp filter, but I think that we shouldn't have unnecessary capabilities in the capability bounding set).

Quackdoc:
well, I doubt it will do anything bad, @aleasto I dont forsee an issue with this getting merged

PR #580, opened 2022-11-10
Mount overlays on top of image mount points

dos1:
This allows the user to make modifications to the images that may persist between image upgrades.

For both the system and vendor image there's a set of two overlays:

- one, specified in config as "overlay", is a read-only persistent overlay meant for stuff like installing privileged apps that should persist between updates
- second one, specified as "overlay_rw", is a read-write overlay that stores the changes made by the user when they remount the mount point in read-write mode. This one is cleaned up when performing image upgrade to not carry potentially incompatible changes on between images

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -119,6 +119,13 @@ def init(args):
             helpers.images.get(args)
         if not os.path.isdir(tools.config.defaults["rootfs"]):
             os.mkdir(tools.config.defaults["rootfs"])
+        if not os.path.isdir(tools.config.defaults["overlay"]):
+            os.mkdir(tools.config.defaults["overlay"])
+            os.mkdir(tools.config.defaults["overlay"]+"/vendor")
+        if not os.path.isdir(tools.config.defaults["overlay_rw"]):
+            os.mkdir(tools.config.defaults["overlay_rw"])
+            os.mkdir(tools.config.defaults["overlay_rw"]+"/system")
+            os.mkdir(tools.config.defaults["overlay_rw"]+"/vendor")
         helpers.lxc.setup_host_perms(args)
         helpers.lxc.set_lxc_config(args)
         helpers.lxc.make_base_props(args)
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -117,6 +117,8 @@ def init(args):
         helpers.images.umount_rootfs(args)
         if args.images_path not in tools.config.defaults["preinstalled_images_paths"]:
             helpers.images.get(args)
+        else:
+            helpers.images.remove_overlay(args)
         if not os.path.isdir(tools.config.defaults["rootfs"]):
             os.mkdir(tools.config.defaults["rootfs"])
         if not os.path.isdir(tools.config.defaults["overlay"]):
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -40,6 +40,9 @@ defaults = {
 }
 defaults["images_path"] = defaults["work"] + "/images"
 defaults["rootfs"] = defaults["work"] + "/rootfs"
+defaults["overlay"] = defaults["work"] + "/overlay"
+defaults["overlay_rw"] = defaults["work"] + "/overlay_rw"
+defaults["overlay_work"] = defaults["work"] + "/overlay_work"
 defaults["data"] = defaults["work"] + "/data"
 defaults["lxc"] = defaults["work"] + "/lxc"
 defaults["host_perms"] = defaults["work"] + "/host-permissions"
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -21,7 +21,8 @@ config_keys = ["arch",
                "vendor_type",
                "system_datetime",
                "vendor_datetime",
-               "suspend_action"]
+               "suspend_action",
+               "mount_overlays"]
 
 # Config file/commandline default values
 # $WORK gets replaced with the actual value for args.work (which may be
@@ -36,7 +37,8 @@ defaults = {
         "/etc/waydroid-extra/images",
         "/usr/share/waydroid-extra/images",
     ],
-    "suspend_action": "freeze"
+    "suspend_action": "freeze",
+    "mount_overlays": "True",
 }
 defaults["images_path"] = defaults["work"] + "/images"
 defaults["rootfs"] = defaults["work"] + "/rootfs"
--- a/tools/helpers/images.py
+++ b/tools/helpers/images.py
@@ -128,8 +128,19 @@ def make_prop(args, cfg, full_props_path):
 def mount_rootfs(args, images_dir, session):
     helpers.mount.mount(args, images_dir + "/system.img",
                         tools.config.defaults["rootfs"], umount=True)
+    helpers.mount.mount_overlay(args, [tools.config.defaults["overlay"],
+                                       tools.config.defaults["rootfs"]],
+                                tools.config.defaults["rootfs"],
+                                upper_dir=tools.config.defaults["overlay_rw"] + "/system",
+                                work_dir=tools.config.defaults["overlay_work"] + "/system")
     helpers.mount.mount(args, images_dir + "/vendor.img",
                            tools.config.defaults["rootfs"] + "/vendor")
+    helpers.mount.mount_overlay(args, [tools.config.defaults["overlay"] + "/vendor",
+                                       tools.config.defaults["rootfs"] + "/vendor"],
+                                tools.config.defaults["rootfs"] + "/vendor",
+                                upper_dir=tools.config.defaults["overlay_rw"] + "/vendor",
+                                work_dir=tools.config.defaults["overlay_work"] + "/vendor")
+
     for egl_path in ["/vendor/lib/egl", "/vendor/lib64/egl"]:
         if os.path.isdir(egl_path):
             helpers.mount.bind(
--- a/tools/helpers/images.py
+++ b/tools/helpers/images.py
@@ -4,6 +4,7 @@ import logging
 import zipfile
 import json
 import hashlib
+import shutil
 import os
 import tools.config
 from tools import helpers
@@ -78,6 +79,7 @@ def get(args):
             tools.config.save(args, cfg)
             os.remove(images_zip)
             break
+    remove_overlay(args)
 
 def replace(args, system_zip, system_time, vendor_zip, vendor_time):
     cfg = tools.config.load(args)
@@ -92,6 +94,13 @@ def replace(args, system_zip, system_time, vendor_zip, vendor_time):
             zip_ref.extractall(args.images_path)
         cfg["waydroid"]["vendor_datetime"] = str(vendor_time)
         tools.config.save(args, cfg)
+    remove_overlay(args)
+
+def remove_overlay(args):
+    if os.path.isdir(tools.config.defaults["overlay_rw"]):
+        shutil.rmtree(tools.config.defaults["overlay_rw"])
+    if os.path.isdir(tools.config.defaults["overlay_work"]):
+        shutil.rmtree(tools.config.defaults["overlay_work"])
 
 def make_prop(args, cfg, full_props_path):
     if not os.path.isfile(args.work + "/waydroid_base.prop"):
--- a/tools/helpers/images.py
+++ b/tools/helpers/images.py
@@ -135,20 +135,23 @@ def make_prop(args, cfg, full_props_path):
     os.chmod(full_props_path, 0o644)
 
 def mount_rootfs(args, images_dir, session):
+    cfg = tools.config.load(args)
     helpers.mount.mount(args, images_dir + "/system.img",
                         tools.config.defaults["rootfs"], umount=True)
-    helpers.mount.mount_overlay(args, [tools.config.defaults["overlay"],
-                                       tools.config.defaults["rootfs"]],
-                                tools.config.defaults["rootfs"],
-                                upper_dir=tools.config.defaults["overlay_rw"] + "/system",
-                                work_dir=tools.config.defaults["overlay_work"] + "/system")
+    if cfg["waydroid"]["mount_overlays"] == "True":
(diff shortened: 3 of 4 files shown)


Quackdoc:
```
(008989) [11:42:01] % mount -o ro /var/lib/waydroid/images/system.img /var/lib/waydroid/rootfs
(008989) [11:42:01] % mkdir -p /var/lib/waydroid/overlay_work/system
(008989) [11:42:01] % mount -t overlay -o ro,lowerdir=/var/lib/waydroid/overlay:/var/lib/waydroid/rootfs,upperdir=/var/lib/waydroid/overlay_rw,workdir=/var/lib/waydroid/overlay_work/system overlay /var/lib/waydroid/rootfs
(008989) [11:42:01] % mount -o ro /var/lib/waydroid/images/vendor.img /var/lib/waydroid/rootfs/vendor
(008989) [11:42:01] % mkdir -p /var/lib/waydroid/overlay_work/vendor
(008989) [11:42:01] % mount -t overlay -o ro,lowerdir=/var/lib/waydroid/overlay/vendor:/var/lib/waydroid/rootfs/vendor,upperdir=/var/lib/waydroid/overlay_rw/vendor,workdir=/var/lib/waydroid/overlay_work/vendor overlay /var/lib/waydroid/rootfs/vendor
mount: /var/lib/waydroid/rootfs/vendor: overlay already mounted on /var/lib/waydroid/rootfs.
       dmesg(1) may have more information after failed mount system call.
```

tested with a fresh init testing with a clean install via 
```
sudo rm -rf /var/lib/waydroid /home/.waydroid ~/waydroid ~/.share/waydroid ~/.local/share/waydroid /usr/share/waydroid-extra/images
```

dos1:
@Quackdoc Which kernel version are you running?

Quackdoc:
> @Quackdoc Which kernel version are you running?

6.0.8

dos1:
@Quackdoc Does this patch help?
```
diff --git a/tools/helpers/mount.py b/tools/helpers/mount.py
index 5b4a7865ed48..0e31fbb123bc 100644
--- a/tools/helpers/mount.py
+++ b/tools/helpers/mount.py
@@ -172,5 +172,5 @@ def mount_overlay(args, lower_dirs, destination, upper_dir=None, work_dir=None,
                 raise RuntimeError("Mount failed, folder does not exist: " +
                                    dir_path)
 
-    mount(args, "overlay", destination, mount_type="overlay", options=options,
+    mount(args, destination, destination, mount_type="overlay", options=options,
           readonly=readonly, create_folders=create_folders, force=True)
```

Which distro are you using, btw?

Quackdoc:
no change, using arch

dos1:
Looks like on my system it defaulted to `index=off`, which is why it worked!

Thanks for testing, should be fixed now :)

Quackdoc:
looks good, would be nice to see some docs written up for this, but other then that I have no complaints, tested and working as intended on arch

ChrysoliteAzalea:
I've installed Waydroid with this patch, and it works fine for me. For example, creating a text file in /system results in placing the file in the ```overlay_rw``` and not in the image itself, which allows for safe modification.

aleasto:
Is it not an issue that /var/lib/waydroid/overlay/vendor is both a directory under the 'system' overlay and its own overlay?

And why does `overlay_rw` instead use `./system` and `./vendor`?

dos1:
> Is it not an issue that /var/lib/waydroid/overlay/vendor is both a directory under the 'system' overlay and its own overlay?

It's only ever used as a lowerdir, so no. I opted to make it closer to actual filesystem structure for ease of use, since the only way to use `/var/lib/waydroid/overlay` is to put things there manually.

> And why does overlay_rw instead use ./system and ./vendor?

That one's an upperdir, and as it turned out in comments above, having them share some paths can cause troubles. Since you don't have to put things there manually but can remount the overlay in rw mode instead, I decided that having them separate is not an issue.

tools/actions/upgrader.py
```diff
@@ -29,6 +29,7 @@ def upgrade(args):
             helpers.images.get(args)
         else:
             logging.info("Upgrade refused because a pre-installed image is detected at {}.".format(args.images_path))
+    helpers.images.remove_overlay(args)
```
aleasto, line 32: `waydroid upgrade --offline` is called on a package update, in debian and fedora at least, to refresh the configs in /var/lib/waydroid. So this would remove the rw overlays for no reason. It should probably only be executed if actually replacing the images.
On the other hand, distros that ship android images through their package manager like Arch would have to remove the overlays when updating the images. I'm not sure if they currently call `waydroid upgrade` or `waydroid upgrade --offline` at all.

Furthermore, `waydroid init --force` should also reset the rw overlays. And maybe even the lower overlays, depending on the intended semantics of `--force`

ChrysoliteAzalea:
This patch seems to cause problems with sound on my computer (in Waydroid only -- on the host namespace, the sound plays fine). The sound also plays just fine on "clean" Waydroid, without this patch.

[logcat.txt](https://github.com/waydroid/waydroid/files/10075029/logcat.txt)

ChrysoliteAzalea:
> @ChrysoliteAzalea Are you using 32-bit Android image / apps? The only way for this PR to influence audio that I can think of would be [alsa-project/alsa-lib#223](https://github.com/alsa-project/alsa-lib/pull/223)

I'm using regular Android image, however I haven't checked whether it's 32-bit or 64-bit. The sound only stops working inside the container (and that affects some apps that play it), but nothing breaks on the host system.

Quackdoc:
I can verify the audio issue happening on my end too after explicitly testing for it. I'm wondering if the overlay is interfering with access to the audio hal stuff as editing out vendor mounting seems to make audio work again.

ChrysoliteAzalea:
I wonder, whether the vendor partition matters that much, or we can only do it for the system partition.

ChrysoliteAzalea:
I also wonder, whether we can simply use Magisk for such tasks, instead of modifying Waydroid itself.

Quackdoc:
I have a mod and maybe another one that would use vendor, and having the overlay system is certainly preferable.

ChrysoliteAzalea:
I wonder, can we simply use Magisk in order to modify the filesystem "systemlessly" (keeping the images intact)?

dos1:
> I'm wondering if the overlay is interfering with access to the audio hal stuff

It really sounds like missing patch from https://github.com/alsa-project/alsa-lib/pull/223, as AFAIK overlayfs inodes are 64-bit - so 32-bit libasound isn't able to read config files, which are necessary for it to even attempt to use the pulse plugin. Which is why I'd like to know whether the affected applications are 32-bit, cause if that's actually the case, then it should be fixable by simply updating ALSA within the image.

Quackdoc:
I tested x64 and arm64 apps to no avail, I dont have the logs right now but I can grab them later.

aleasto:
It doesn't matter the app arch because audio calls go through binder to the audio service, which in facts is 32bit: https://cs.android.com/android/platform/superproject/+/android-11.0.0_r1:hardware/interfaces/audio/common/all-versions/default/service/Android.bp;l=11

dos1:
OK, so I believe updating alsa-lib to at least 1.2.7.1 should fix the audio issue. If that's too much of a hassle for some reason (looks like it's currently at 1.1.8), simply cherry-picking the patch from MR I linked above should be enough.

dos1:
Magisk is a much more complicated way of achieving a similar (although limited) thing. I had a quick look at it and I don't think it's going to work without putting some effort into it, which doesn't really make much sense when you can simply use overlayfs instead.

aleasto:
> OK, so I believe updating alsa-lib to at least 1.2.7.1 should fix the audio issue. If that's too much of a hassle for some reason (looks like it's currently at 1.1.8), simply cherry-picking the patch from MR I linked above should be enough.

Ok, but now I'm thinking if this is causing issues for other 32bit services

Quackdoc:
the migration

> > I'm wondering if the overlay is interfering with access to the audio hal stuff
> 
> It really sounds like missing patch from [alsa-project/alsa-lib#223](https://github.com/alsa-project/alsa-lib/pull/223), as AFAIK overlayfs inodes are 64-bit - so 32-bit libasound isn't able to read config files, which are necessary for it to even attempt to use the pulse plugin. Which is why I'd like to know whether the affected applications are 32-bit, cause if that's actually the case, then it should be fixable by simply updating ALSA within the image.

Assuming this is the case we can probably mount using `xino=off` and **maybe** get this to work, needs testing, and I cant do it @ChrysoliteAzalea maybe you can test? or maybe it's xino=on? not too sure, you would add it to the mounting arguments using -o xino=off

dos1:
> Ok, but now I'm thinking if this is causing issues for other 32bit services

It shouldn't, and if it does it's a clear bug in that service that can make things break in other circumstances as well. I made the above-mentioned ALSA patch a few months ago to fix 32-bit ALSA games not working on Steam Deck - funny that it pops up here as well.

Also, LFS has been a thing since 1990s. ALSA could (barely) be excused for still carrying some old code that doesn't handle it, but Android can't ;)

`xino` may workaround the issue, but I'd rather fix ALSA - there's been already a few stable releases containing the fix out there.

Quackdoc:
> `xino` may workaround the issue, but I'd rather fix ALSA - there's been already a few stable releases containing the fix out there.

rather then it being a work around, I thought it was common practice when you needed 32bit app support, the help text for `CONFIG_OVERLAY_FS_XINO_AUTO` even reccomends disabling it for when 32bit app support is necessary. 

well in either case, I think either solution would be acceptable, we can always add a note to the docs in the debugging section. I have also confirmed that adding xino=off does indeed work at the very least, so thats a route forwards anyways. if it turns out other apps are effected.

aleasto:
Only fixing ALSA in the vendor image means that this commit will depend on all users updating the vendor image.
If there's something we can do on the python side, I'm much more comfortable merging this

aleasto:
Mh, I've picked the alsa fixes but can still reproduce the issue by setting `xino=auto` (it's off by default in my kconfig).

They didn't apply clean so it's possible i have missed something in this older version of alsa. Updating alsa is probably a big task since the android build system is not implemented upstream.

I will try some more but definitely +1 on `xino=off`

jpumc:
Hi,

I have experimented with waydroid and overlays a bit, but I had a problem with overlayfs being picky with the filesystems of the layer dirs. My root fs is ext4 with casefolding enabled which overlayfs refused to use. I worked around this by mounting a image in a directory and then using that directory in the overlay. Could this be supported by this feature?

I found some discussion about a similar issue in https://lkml.kernel.org/linux-ext4/85k1e39qeq.fsf@collabora.com/T/#m3a18331ba3d0dd60b5f4651109440d5d320f6aa5 with some relevant part below

> > [...], what other filesystems are (not) supported?
> 
> There are some special cases like /proc/sys and auto mount points,
> but the most common reason for unsupported lower is case insensitive filesystems
> and filesystems that support unicode character folding.
> Those filesystems MAY be case insensitive/unicode, depending on
> mkfs/mount options:
>  adfs affs cifs efivarfs fat hfs hfsplus hpfs isofs isofs jfs xfs(*) ext4(**)
> 
> (*) xfs case insensitive-ness feature is hidden from dcache (dcache is
> disabled), so
> overlayfs mount won't fail, but it may have unexpected behaviors.
> 
> (**) ext4 supports per directory case insensitive and unicode folding
> since v5.2-rc1,
> if the filesystem was configured that way with mkfs/tune2fs. In this
> case, regardless of
> whether the lower dir is case insensitive or not, overlayfs mount will fail.

Quackdoc:
@jpumc can you elaborate more on the setup you used for mounting?

ChrysoliteAzalea:
> Magisk is a much more complicated way of achieving a similar (although limited) thing. I had a quick look at it and I don't think it's going to work without putting some effort into it, which doesn't really make much sense when you can simply use overlayfs instead.

As far as I know, Magisk also offers various additional functions, such as Zygisk that allows to hide root (I've noticed that some apps detect root even on stock, non-rooted Waydroid images).

Quackdoc:
> @Quackdoc so the way I did the overlay was to create a overlay image file (lets say "overlay.img" with squashfs filesystem, but any fs supported by overlayfs should work). Then before mounting the rootfs overlayfs this "overlay.img" is mounted into the "overlay" dir. The rootfs overlayfs can then be mounted as if "overlay" was a normal dir.

I would be worried about some of the issues with xino, but it could be worth testing for an optional change, folder is prefered by default since it's a lot more simple to work with

aleasto:
So the only issue really is when lowerdir has casefolding.
Do we want to support that case? I'd say it's quite extreme, but maybe there's a painless way?

nitanmarcel:
When `overlay` is mounted, any changes made from inside `waydroid` will be written to `overlay_rw` and removed after OTAs.

I'm curious if that's intended.

aleasto:
of course

aleasto:
Addressed my comments, added `xino=off` and made it possible to opt out since we might not be able to guarantee mounting overlayfs to work everywhere

aleasto:
`init -f` now removes overlays even for preinstalled images

ayasa520:
Would it be possible to improve the current mounting method? `/var/lib/overlay_rw/system` is overlaid as upperdir onto `/var/lib/rootfs/`. This means that after remounting rootfs as read-write, modifications to system will be written to `/var/lib/overlay_rw/system/system` .

https://github.com/waydroid/waydroid/pull/580/files#diff-309130090d80ed7b1d062cbf8243911dac71a786bf8db18e06ceb13735a13b8aR145

aleasto:
This is exactly intended.

ayasa520:
In the `overlay` directory, `system` maps to `/system` and `vendor` maps to `/vendor`. However, in `overlay_rw`, `system/system` maps to `/system` and `vendor` maps to `/vendor`. So the `system` and `vendor` in `overlay_rw` actually refer to `system.img` and `vendor.img`?

aleasto:
That's right.
See [this](https://github.com/waydroid/waydroid/pull/580#issuecomment-1321352012) comment for the two overlay directory structures differ.

PR #662, opened 2022-12-29
Dbus

aleasto:
Reasoning:
* No more polling the file system
* No more need for a global read-writable directory
* Service activation on signal, also optionally integrated with systemd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ APPS_DIR := $(PREFIX)/share/applications
 METAINFO_DIR := $(PREFIX)/share/metainfo
 SYSD_DIR := $(PREFIX)/lib/systemd/system
 DBUS_DIR := $(PREFIX)/share/dbus-1
+POLKIT_DIR := $(PREFIX)/share/polkit-1
 APPARMOR_DIR := /etc/apparmor.d
 
 INSTALL_WAYDROID_DIR := $(DESTDIR)$(WAYDROID_DIR)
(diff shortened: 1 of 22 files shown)


aleasto:
Reminder to self: add dbus dependency to debian

ChrysoliteAzalea:
I'd like to add, that on Ubuntu and systems that use Ubuntu-patched kernel, D-Bus is restricted by AppArmor, so there could be a need for AppArmor policy amendment.

aleasto:
Isn't apparmor limited to processes inside the container?

ChrysoliteAzalea:
> Isn't apparmor limited to processes inside the container?

AFAIK, yes, since, the profile is applied by LXC. If D-Bus is going to be used to processes outside of the container, and these processes are supposed to be unconfined, then no policy amendment would be needed (and I doubt that Android processes are going to use D-Bus anyway).

JamiKettunen:
@CajuM If you're still the NixOS Waydroid package maintainer can you check how the latest `main` branch commit (perhaps with #662 included) works? v1.4.0 is near release which includes D-Bus activation by default.

One potential issue I see is [`dbus/id.waydro.Container.service`](https://github.com/waydroid/waydroid/blob/main/dbus/id.waydro.Container.service) hardcoding `Exec=/usr/bin/waydroid -w container start` for the binary path which might need adaptation, just asking so it can perhaps be addressed on the Makefile side with a variable for the path or similar.

PR #670, opened 2023-01-02
net: Use custom port for dnsmasq

aleasto:


aleasto:
@Quackdoc does this look right

Quackdoc:
im on mobile so no indepth check, but port 5353 might pose troublesome, a good amount of services use it.

notramo:
Why is it necessary?

aleasto:
I don't know how to tell android to use this port

Issue #661, opened 2022-12-28
`ashmem` module problem


notramo:
`waydroid session start` complains about `ashmem` not existing.

PinePhone, postmarketos v22.12
Linux 6.1.0-rc7
waydroid 1.3.3
`sys.use_memfd=true` option is set in `/var/lib/waydroid/waydroid.prop` and `/var/lib/waydroid/waydroid-base.prop`

```
(010852) [15:49:21] Downloading https://sourceforge.net/projects/waydroid/files/images/system/lineage/waydroid_arm64/lineage-18.1-20221224-VANILLA-waydroid_arm64-system.zip/download
(010852) [16:19:41] Validating system image
(010852) [16:19:55] Extracting to /var/lib/waydroid/images
(010852) [16:21:33] Save config: /var/lib/waydroid/waydroid.cfg
(010852) [16:21:38] Downloading https://sourceforge.net/projects/waydroid/files/images/vendor/waydroid_arm64/lineage-18.1-20221224-MAINLINE-waydroid_arm64-vendor.zip/download
(010852) [16:23:37] Validating vendor image
(010852) [16:23:38] Extracting to /var/lib/waydroid/images
(010852) [16:23:50] Save config: /var/lib/waydroid/waydroid.cfg
(010852) [16:23:50] % lxc-info --version
5.0.1
(010852) [16:23:50] % mkdir -p /var/lib/waydroid/lxc/waydroid
(010852) [16:23:50] % sh -c cat "/usr/lib/waydroid/data/configs/config_base" "/usr/lib/waydroid/data/configs/config_2" "/usr/lib/waydroid/data/configs/config_4" > "/var/lib/waydroid/lxc/waydroid/config"
(010852) [16:23:50] % sed -i s/LXCARCH/aarch64/ /var/lib/waydroid/lxc/waydroid/config
(010852) [16:23:50] % cp -fpr /usr/lib/waydroid/data/configs/waydroid.seccomp /var/lib/waydroid/lxc/waydroid/waydroid.seccomp
(010852) [16:23:50] % mv /var/lib/waydroid/config_nodes /var/lib/waydroid/lxc/waydroid
(016149) [16:27:09] Save session config: /var/lib/waydroid/session.cfg
(016149) [16:27:09] UserMonitor service is not even started
(016149) [16:27:09] Clipboard service is not even started
(016682) [16:27:53] % modprobe -q ashmem_linux
(016682) [16:27:53] % chmod 666 -R /dev/binder
[...]

PR #672, opened 2023-01-04
lxc: Only add /dev/ashmem to config_nodes when it exists

JamiKettunen:
This hasn't existed in mainline Linux kernels [since v5.18](https://cateee.net/lkddb/web-lkddb/ASHMEM.html) and trying to always mount it regardless just adds to the noise in `waydroid log` in most cases, so simply don't add it to `config_nodes` unless it actually exists.

Silences the following:
```
lxc-start: waydroid: ../src/lxc/utils.c: safe_mount: 1221 No such file or directory - Failed to mount "/dev/ashmem" onto "/usr/lib/lxc/rootfs/dev/ashmem"
```

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -128,6 +128,7 @@ def init(args):
             os.mkdir(tools.config.defaults["overlay_rw"])
             os.mkdir(tools.config.defaults["overlay_rw"]+"/system")
             os.mkdir(tools.config.defaults["overlay_rw"]+"/vendor")
+        helpers.drivers.probeAshmemDriver(args)
         helpers.lxc.setup_host_perms(args)
         helpers.lxc.set_lxc_config(args)
         helpers.lxc.make_base_props(args)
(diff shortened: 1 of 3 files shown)


cniw:
Fixes: https://github.com/waydroid/waydroid/issues/661

JamiKettunen:
Hm I now wonder if `helpers.drivers.loadBinderNodes(args)` should be called before `make_entry("/dev/ashmem")` so it could be potentially picked up during the config creation in case the module wasn't already probed, or if there is any other way this could be silenced on systems where  `ashmem` is irrelevant

aleasto:
> Hm I now wonder if `helpers.drivers.probeAshmemDriver(args)` should be called before `make_entry("/dev/ashmem")` so it could be potentially picked up during the config creation in case the module wasn't already probed

Yeah do that.
Discussion in AOSP seems to be moving towards enabling `sys.use_memfd` by default but also keep ashmem to support applications that use `/dev/ashmem` directly

aleasto:
works

PR #675, opened 2023-01-05
configs: Drop CAP_WAKE_ALARM capability

aleasto:
Prevent the android userspace from continuously waking up the hardware


aleasto:
@dos1 are you sure this may causes crashes?

aleasto:
yep. definitely crashes

Issue #589, opened 2022-11-13
No 3D rendering in Sky: Children of the Light using AMD RDNA2 GPU


FridgeRacer:
**Describe the bug**
When trying to play Sky: Children of the Light, only 2D elements (text) are rendered, everything behind (which should be 3D elements) is completely black.

**Screenshots**
![Sky: Children of the Light bug](https://user-images.githubusercontent.com/40672277/201539718-6f436c41-4811-416a-b4e8-1bb53ce65c28.png)

**General information:**
 - Waydroid tools Version:  1.3.3.r3.ge074cf4-1 (manually built)
 - Waydroid Images Version:  GAPPS 20221108 (Vanilla version also tested)

**Desktop:**
 - OS: Arch Linux
 - CPU: AMD Ryzen 5 2600
 - GPU: AMD Radeon RX 6700 XT (Open source driver + RADV)
 - Kernel version: 6.0.8
 - Host mesa version: Mesa 22.2.3
 - Desktop: KDE Plasma 5.26.3

**Additional context**
- The game is proven to work fine with an Intel GPU. I am also using libhoudini to translate aarch64 to x86_64.
By monitoring the Waydroid log with tail -f, I don't see any errors when launching the game.
- There are no error messages in the game or the Android environment either, however the framerate of the text that does appear in game is rather low.
- When idling on the screen shown above, GPU utilization is close to nil, there is also very little CPU usage.
- Pressing "Sign In Method" should bring up a menu of ways to sign in to the game, but it does not appear even if the button is being registered as tapped by the game.
- By pressing play, I can hear the sound from the game and that it is continuing to the new player tutorial, but it still doesn't render graphics other than 2D elements (tutorial text).

**Logs**
 - Prop file: [waydroid_base.prop.txt](https://github.com/waydroid/waydroid/files/9998024/waydroid_base.prop.txt)
 - Waydroid log: [waydroid.log.txt](https://github.com/waydroid/waydroid/files/9998026/waydroid.log.txt)
 - Waydroid logcat: [waydroid_logcat.txt](https://github.com/waydroid/waydroid/files/9998045/waydroid_logcat.txt)

Quackdoc:
change `ro.hardware.gralloc=gbm` to `ro.hardware.gralloc=minigbm_gbm_mesa` amd has a known vulkan issue that should be resolved by this. good chance this will fix it

FridgeRacer:
[...]

PR #676, opened 2023-01-05
gpu: Also mount card node

aleasto:
Necessary for minigbm_gbm_mesa


aleasto:
I can't remember why i hadn't pushed this

JamiKettunen:
Cool, what does `minigbm_gbm_mesa` bring us? I saw at least one issue about it making some game work: #589

Quackdoc:
does it need it? works fine for me on AMD RX580 without it. is this an intel gpu thing? 

Either way no complaints from me if it's necessary.

aleasto:
https://github.com/waydroid/android_external_minigbm/blob/lineage-18.1/gbm_mesa_driver/gbm_mesa_internals.cpp#L204

An explanation can be found here: https://github.com/robherring/gbm_gralloc/pull/16

Quackdoc:
ahh I see,

PR #677, opened 2023-01-06
Fix prop set command

ZhaoZuohong:
There is a small typo that misses `tools.` when calling `helpers`, which has issues with setting props.


aleasto:
Oops, thanks

PR #678, opened 2023-01-06
add a dbus activated systemd user service for waydroid session

Fuseteam:
this will start waydroid session over dbus

currently it reaches https://github.com/waydroid/waydroid/blob/fb92d3a016a956e8520e938b59f55bd0fe9c2deb/tools/actions/app_manager.py#L65 too quickly causing it to return `Failed to get service waydroidplatform, trying again...` on the first run, it launches fine on the second

my current understanding is that it should  only call `launchNow` after this point has been reached https://github.com/waydroid/waydroid/blob/fb92d3a016a956e8520e938b59f55bd0fe9c2deb/tools/services/user_manager.py#L75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ WAYDROID_DIR := $(PREFIX)/lib/waydroid
 BIN_DIR := $(PREFIX)/bin
 APPS_DIR := $(PREFIX)/share/applications
 METAINFO_DIR := $(PREFIX)/share/metainfo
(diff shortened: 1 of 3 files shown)


JamiKettunen:
Rebase please instead of merging `main` back to your PR branch when you want to update it ^^ Not sure if this repo follows semi-linear git history but it's good practice either way in general

Fuseteam:
yeah will do, i thought the github interface would do the sane thing

PR #691, opened 2023-01-11
Recursively mount data directory

aleasto:
A user may set up "shared folders" under ~/.local/share/waydroid/data


aleasto:
superseeded

Issue #693, opened 2023-01-15
Prevent session start without existing Wayland socket


JamiKettunen:
(New) users keep launching the session as `root` due to prefixing *literally every* command with `sudo` even when not asked to, or launching Waydroid under X11 without any existing Wayland socket under `XDG_RUNTIME_DIR`.

Preventing session launch as root was the primary goal of https://github.com/waydroid/waydroid/pull/246 which has stalled for now, so this would also supersede that.

PR #696, opened 2023-01-15
session: Add strict WAYLAND_DISPLAY validity checks

JamiKettunen:
Enough with not checking we actually have a Wayland compositor around at all: start requiring the `WAYLAND_DISPLAY` socket actually exists as an absolute path or relatively under `XDG_RUNTIME_DIR`.

Additionally if `WAYLAND_DISPLAY` isn't an absolute path to the socket (most setups) ensure `XDG_RUNTIME_DIR` is set and error with a typically appropriate message.

Fixes https://github.com/waydroid/waydroid/issues/693, closes https://github.com/waydroid/waydroid/pull/246.

Also contains a few small irrelevant commits that shouldn't change runtime in any way.

Few questions though (@aleasto):
1. Should we support `WAYLAND_SOCKET` right now? [technically it's a part of the spec](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1af048371dfef7577bd39a3c04b78d0374) so I left it as a `TODO` for now
2. Should we add a default for `XDG_RUNTIME_DIR` (of e.g. `/run/user/$(id -u)`)? `sudo waydroid show-full-ui` currently produces this and I kinda like how obvious it makes that the user attempted to start a session as root without any `XDG_RUNTIME_DIR` configured (though perhaps that could be another check by itself? I do like that idea as well):
```
[06:19:14] Starting waydroid session
[06:19:14] WAYLAND_DISPLAY is not set, defaulting to "wayland-0"
[06:19:14] WAYLAND_DISPLAY socket 'None/wayland-0' doesn't exist; you're running a Wayland compositor right?
```

### DIFF
--- a/tools/actions/session_manager.py
+++ b/tools/actions/session_manager.py
@@ -39,13 +39,30 @@ def start(args, unlocked_cb=None):
             unlocked_cb()
         return
 
-    session = copy.copy(tools.config.session_defaults);
+    session = copy.copy(tools.config.session_defaults)
+
+    # TODO: also support WAYLAND_SOCKET?
     wayland_display = session["wayland_display"]
     if wayland_display == "None" or not wayland_display:
         logging.warning('WAYLAND_DISPLAY is not set, defaulting to "wayland-0"')
+        wayland_display = session["wayland_display"] = "wayland-0"
+
+    if os.path.isabs(wayland_display):
+        wayland_socket_path = wayland_display
+    else:
+        xdg_runtime_dir = session["xdg_runtime_dir"]
+        if xdg_runtime_dir == "None" or not xdg_runtime_dir:
+            logging.error(f"XDG_RUNTIME_DIR is not set; please don't start a Waydroid session with 'sudo'!")
+            sys.exit(1)
+        wayland_socket_path = os.path.join(xdg_runtime_dir, wayland_display)
+    if not os.path.exists(wayland_socket_path):
+        logging.error(f"Wayland socket '{wayland_socket_path}' doesn't exist; are you running a Wayland compositor?")
+        sys.exit(1)
+
     waydroid_data = session["waydroid_data"]
     if not os.path.isdir(waydroid_data):
         os.makedirs(waydroid_data)
+
     dpi = tools.helpers.props.host_get(args, "ro.sf.lcd_density")
     if dpi == "":
         dpi = os.getenv("GRID_UNIT_PX")
--- a/debian/waydroid.postinst
+++ b/debian/waydroid.postinst
@@ -3,7 +3,7 @@ set -e
 
 configure() {
 	# Update configs
-    waydroid upgrade -o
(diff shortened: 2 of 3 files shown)


JamiKettunen:
Hm I guess another question is: was absolute `WAYLAND_DISPLAY` path to the socket supported even before this PR in general?

aleasto:
> 1. Should we support `WAYLAND_SOCKET` right now? [technically it's a part of the spec](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1af048371dfef7577bd39a3c04b78d0374) so I left it as a `TODO` for now

There's no way we're going to pass an open file descriptor to the container. So it would involve retrieving back the name from `/proc/self/fd/$WAYLAND_SOCKET` and setting WAYLAND_DISPLAY accordingly: meh.

> 2. Should we add a default for `XDG_RUNTIME_DIR` (of e.g. `/run/user/$(id -u)`)? `sudo waydroid show-full-ui` currently produces this and I kinda like how obvious it makes that the user attempted to start a session as root without any `XDG_RUNTIME_DIR` configured (though perhaps that could be another check by itself? I do like that idea as well):
> 
> ```
> [06:19:14] Starting waydroid session
> [06:19:14] WAYLAND_DISPLAY is not set, defaulting to "wayland-0"
> [06:19:14] WAYLAND_DISPLAY socket 'None/wayland-0' doesn't exist; you're running a Wayland compositor right?
> ```
It's ok.

> Hm I guess another question is: was absolute `WAYLAND_DISPLAY` path to the socket supported even before this PR in general?

It depends on the `libwayland-client` version. In android-11 images it is: https://android.googlesource.com/platform/external/wayland/+/refs/tags/android-11.0.0_r1/src/wayland-client.c#961

tools/actions/session_manager.py
```diff
@@ -39,13 +39,24 @@ def start(args, unlocked_cb=None):
...
+    # TODO: also support WAYLAND_SOCKET?
     wayland_display = session["wayland_display"]
     if wayland_display == "None" or not wayland_display:
         logging.warning('WAYLAND_DISPLAY is not set, defaulting to "wayland-0"')
+        wayland_display = session["wayland_display"] = "wayland-0"
+
+    xdg_runtime_dir = session["xdg_runtime_dir"]
+    wayland_socket = wayland_display if os.path.isabs(wayland_display) else os.path.join(xdg_runtime_dir, wayland_display)
+    if not os.path.exists(wayland_socket):
+        logging.error(f"WAYLAND_DISPLAY socket '{wayland_socket}' doesn't exist; you're running a Wayland compositor right?")
```
aleasto, line 53: `f"Wayland socket '{wayland_socket}' doesn't exist; are you running a Wayland compositor?"`

aleasto:
I think i prefer version 6b49e66b4f5b0493ad550ec02ecc7c7d4fd1396a

Issue #704, opened 2023-01-19
Mount XDG_RUNTIME_DIR dynamically on session startup


JamiKettunen:
Currently `waydroid init` [defines a bunch mount of entries](https://github.com/waydroid/waydroid/blob/1.3.4/tools/helpers/lxc.py#L24) such as `lxc.mount.entry = /run run none rbind,create=dir 0 0` or `lxc.mount.entry = /dev/shm dev/shm none rbind,create=dir,optional 0 0` on `/var/lib/waydroid/lxc/waydroid/config_nodes` blindly because they *might* contain `XDG_RUNTIME_DIR`, but since this is a user configurable variable this is just a massive hack and only covers the most common scenarios pre-setup by distributions out of the box.

For example the following `.bash_profile` will result in Waydroid not launching due to missing Wayland socket in the container as the parent dir won't be bind-mounted over from host:
```bash
#!/bin/bash
if test -z "${XDG_RUNTIME_DIR}"; then
    export XDG_RUNTIME_DIR=/tmp/${UID}-runtime-dir
    if ! test -d "${XDG_RUNTIME_DIR}"; then
        mkdir "${XDG_RUNTIME_DIR}"
        chmod 0700 "${XDG_RUNTIME_DIR}"
    fi
fi
```
Perhaps some LXC `pre-start` script could be used here for mounting? 

@aleasto also suggested:
> you could force it to be at like `/tmp/xdg`, then before starting the session `mount -o bind $XDG_RUNTIME_DIR /tmp/xdg`
> but then the same should apply to `PULSE_RUNTIME_PATH`
> and maybe some other things

and also:
> or, we could have one lxc config file dedicated to session stuff
> included from the main one
> that is generated on session start

For additional context start reading from https://t.me/WayDroid/129454

dchmelik:
What about for Slackware?

JamiKettunen:
@dchmelik what do you mean? you should make a new issue with full details on your actual problem instead of vague comments on a random 2+ year old closed issue that was already fixed way back

PR #709, opened 2023-01-21
lxc: Create session-specific mount entries

aleasto:
Closes: #704


aleasto:
Verified on sailfish

aleasto:
Included `/data` mount in the new mechanism.
We have to keep the old bind mount logic for compatibility because some distributions don't perform `waydroid ugprade --offline` on package upgrade

PR #722, opened 2023-01-30
arch: Fallback to arm on weird arm64 kernels with 32-bit userspace

JamiKettunen:
Typically `armv8l` should catch these, but it seems some Android downstream kernels at least cheat and don't report the correct thing.


GamerSwir2 (commented):
works fine

maciek134:
why not just `platform.architecture()` instead of `sys.maxsize` that's not immediately clear what it does?

aleasto:
> why not just `platform.architecture()` instead of `sys.maxsize` that's not immediately clear what it does?

I prefer that, if it works equally

PR #725, opened 2023-01-30
Switch AppArmor profiles for the Waydroid container into "enforce" mode

ChrysoliteAzalea:
Hello everyone!

During the last year, [a commit has been merged](https://github.com/waydroid/waydroid/pull/506) that enables AppArmor for the container. In order to avoid breaking any existing installations and work out policy mistakes, all profiles have been switched into the *complain* mode, in which AppArmor handles domain transitions, logs violations, but doesn't enforce rules (except for deny rules). If proposed AppArmor policies are somehow too restrictive for a container to work properly, it will show up in the audit log.

If you have AppArmor enabled on your system, you can help with AppArmor support in Waydroid:

* by reporting any policy violations by Android container processes in the audit log on your system (/var/log/audit/audit.log if auditd is installed and system log if it's not). They are reported in the AVC category, with "ALLOWED" label if the profile is in *complain* mode and "DENIED" label if it's in *enforce* mode
* by switching the profiles into *enforce* mode and reporting any problems in Waydroid container resulted from that

### DIFF
--- a/data/configs/apparmor_profiles/adbd
+++ b/data/configs/apparmor_profiles/adbd
@@ -1,4 +1,4 @@
-profile adbd flags=(attach_disconnected,mediate_deleted,complain) {
+profile adbd flags=(attach_disconnected,mediate_deleted) {
   /** ix,
   /dev** rw,
   network,
--- a/data/configs/apparmor_profiles/android_app
+++ b/data/configs/apparmor_profiles/android_app
@@ -1,4 +1,4 @@
-profile android_app flags=(attach_disconnected, complain, mediate_deleted) {
+profile android_app flags=(attach_disconnected, mediate_deleted) {
   /** ix,
   /dev** rw,
   network,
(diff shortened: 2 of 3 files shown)


JamiKettunen:
on Ubuntu Touch 20.04 I already tested AppArmor previously, but on some boots the `lxc-waydroid` profile seemingly just "vanishes" and doesn't load automatically for some odd reason preventing the launch altogether; maybe you'd have some ideas / have seen this in the wild?

Downstream UT issue tracker: https://gitlab.com/ubports/development/core/packaging/waydroid/-/issues/14

ChrysoliteAzalea:
I run Waydroid on a desktop system, not Ubuntu Touch, however, I can try Ubuntu Touch in QEMU to see what's going on. I don't know how the profile loading is organized in Ubuntu Touch and whether it's different from a desktop.

aleasto:
I had to revert this, it breaks starting the GAPPS image (denied read to /data/resource-cache) and dmesg shows even more denials

ChrysoliteAzalea:
Oh sorry, I don't use Google Play Services. I hope that profiles will be switched to "enforce" mode after issues with gapps are worked out.

PR #734, opened 2023-02-02
Update appstream

Doomsdayrs:
The app stream file was not validated via appstream-glib.

This commit ensures that the appstream passes validation.

I suggest checking the content rating, I did it under the assumption of Google.


aleasto:
Thank you. It's tricky to define content rating but i suppose it's ok

PR #787, opened 2023-03-04
Update Makefile to have meaningful targets

mazunki:
The aim of this pull request is to clean up pull request #484, in order to prepare to support multiple service managers.

I have not added any new feature here, I have simply moved some files around, and given names to different targets in the Makefile. The only practical difference is that there will now be a few more files under `/etc/waydroid`, which are sourced from `waydroid-net.sh`. 

I noticed there are now `config_{1,3,4}`, but no `config_2`. Furthermore, it doesn't seem like these are used anywhere. In the previous failed pull request I had done something similar, but before I did anything I'd figure I'd ask what the reasoning behind it was.

After this PR is completed, I'll create a new PR for supporting s6.

- moved configuration into its own file
- moved nft rules into their own files
- split up the makefile into meaningful modules


aleasto:
> I noticed there are now config_{1,3,4}, but no config_2. Furthermore, it doesn't seem like these are used anywhere. In the previous failed pull request I had done something similar, but before I did anything I'd figure I'd ask what the reasoning behind it was.

https://github.com/waydroid/waydroid/blob/main/tools/helpers/lxc.py#L138

mazunki:
> https://github.com/waydroid/waydroid/blob/main/tools/helpers/lxc.py#L138

Thanks, I had missed that.

JamiKettunen:
@aleasto Any further thoughts on this?

JamiKettunen:
Ping @aleasto, especially the network script configuration in `/etc` is rather useful since you wouldn't have to hack the `/usr/lib/waydroid/data/scripts/waydroid-net.sh` which gets replaced on package upgrades etc (and distro policies typically wouldn't allow marking a "configuration file" either way)

PR #817, opened 2023-03-19
Waydroid overwrites user desktop file

Boruch-Baum:
This is actually three bugs: 

1) When I modify my $HOME/.local/share/applications/Waydroid.desktop file, waydroid overwrites it with its default when it next launches.

2) Because of how waydroid is set up, it succeeds in doing so even when that file is chowned root:root !!!

3) When the file is made immutable (chattr +i), waydroid fails to launch, seemingly because it can't modify the file !!!


JamiKettunen:
Making the previous code comments isn't useful, just leaves useless "dead code" in the file that will never be run. There probably should be some `/var/lib/waydroid/waydroid.cfg` item like `rewrite_desktop_file`; see e87b848b9b2acaa4952eb92e339e575d25281d3a for example how `mount_overlays` option was added.

Twig6943:
Could somebody close this as it was solved by https://github.com/waydroid/waydroid/pull/1150 ?

Issue #845, opened 2023-04-07
Is there any way to avoid running adb connect every time?


mitsukuri:
I'm using Waydroid to test a mobile app via appium/uiautomator2.
As it currently stands, I have to issue `adb connect 192.168.240.112:5555` manually at least once before I run my test script.

Is there a way to hardcode this somewhere, e.g. as an appium/webdriverio/uiautomator2 capability, other than running `adb connect` from my script?

mitsukuri:
Of course this is not the right place to ask this question, but I'd be happy if anyone had similar experience and could chime in. Closing.

Issue #164, opened 2021-10-19
Allow interaction with Android Device Bridge


TheOneWithTheBraid:
In the future, it might be helpful to enable support for the Android Device Bridge (adb) to install apps and perform actions like shell commands, as many existing tools rely on it. This might be particularly useful in case you use Waydroid to debug Android apps you develop just on your machine.

rizzini:
+1 vote for this.

afzl-wtu:
I think you can enable wireless debugging and connect through it.

UlyssesZh:
> I think you can enable wireless debugging and connect through it.

Does not work in some cases: #379

aleasto:
You can use adb over tcp: https://docs.waydro.id/faq/using-adb-with-waydroid
I'm not sure if it would be feasible to have it working by emulating the USB interface instead...

Thesola10:
I remember Anbox having had a feature to bridge directly to the local ADB daemon, how did they do that and can Waydroid emulate it?

mak448a:
> You can use adb over tcp: https://docs.waydro.id/faq/using-adb-with-waydroid I'm not sure if it would be feasible to have it working by emulating the USB interface instead...

I know this is late, but do you know how I can use adb over wifi? The waydroid device refuses to enable wifi and "Wireless Debugging" under development options.

7merlin:
I have the same question: it's not possible to do adb over network without this setting and it can only be enabled if there's WiFi. It's on Android 11

thehouseisonfire:
+1 same problem with enabling Wifi (and thus wireless debugging)

PR #847, opened 2023-04-10
Add device IP address to status, auto adb connect

dasisdormax:
To make my workflow during android development easier, I implemented
- a display of Waydroid's local IP address to the status command
- an automatic call to ADB connect (if available on the host system) once the system is ready

The IP address is taken from the DHCP lease file for the default network configuration. Tested on cachyos with Waydroid installed from the AUR.

As I am not very familiar with Python or with the code structure, any feedback is very appreciated. Please let me know if this fits the project or if changes are necessary for issues or use cases that I didn't think of.

Possible alternatives:
- Just displaying the IP address in the status
- Adding a subcommand, such as `waydroid adb connect` instead of automatically connecting

Some related issues: #164 #845

### DIFF
--- a/tools/helpers/net.py
+++ b/tools/helpers/net.py
@@ -1,0 +1,36 @@
+# Copyright 2023 Maximilian Wende
+# SPDX-License-Identifier: GPL-3.0-or-later
+from shutil import which
+import tools.helpers.run
+import logging
+import re
+
+def adb_connect(args):
+    """
+    Creates an android debugging connection from the host system to the
+    Waydroid device, if ADB is found on the host system and the device
+    has booted.
+    """
+    # Check if adb exists on the system.
+    if not which("adb"):
+        return
+
+    # Start and 'warm up' the adb server
+    tools.helpers.run.user(args, ["adb", "start-server"])
+
+    ip = get_device_ip_address()
+    if not ip:
+        return
+
+    tools.helpers.run.user(args, ["adb", "connect", ip])
+    logging.info("Established ADB connection to Waydroid device at {}.".format(ip))
+
+def get_device_ip_address():
+    # The IP address is queried from the DHCP lease file.
+    lease_file = "/var/lib/misc/dnsmasq.waydroid0.leases"
+
+    try:
+        with open(lease_file) as f:
+            return re.search("(\d{1,3}\.){3}\d{1,3}\s", f.read()).group().strip()
+    except:
+        pass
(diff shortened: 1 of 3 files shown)


tools/helpers/net.py
```diff
@@ -0,0 +1,36 @@
...
+    Creates an android debugging connection from the host system to the
+    Waydroid device, if ADB is found on the host system and the device
+    has booted.
+    """
+    # Check if adb exists on the system.
+    if not which("adb"):
+        return
+
+    # Start and 'warm up' the adb server
+    tools.helpers.run.user(args, ["adb", "devices"])
```
aleasto, line 19: `adb start-server` would be the intended way

aleasto (commented):
Implementation looks good.

I think i would be personally annoyed to have an ad device always connected (waydroid is also designed to sit quietly in the background) but for the majority it's probably a positive change.

Can you verify this doesn't break too badly when the LXC container freezes? Waydroid will automatically freeze after a short timeout when no Wayland windows are open, or you may manually freeze with `sudo waydroid container freeze`

tools/helpers/net.py
dasisdormax: Thanks, I updated my branch

dasisdormax:
With the container freezed, the following happened in my testing:
- The device appears to be still connected in `adb devices` or Android Studio
- Commands such as `adb shell` will freeze and Android Studio commands (run, debug, etc.) will just wait for the device. These commands however resume once the container unfreezes.

Apart from that, the ADB connection was lost occasionally, but I couldn't find an exact cause and it happened independently of freezing / unfreezing of the container.

For me, this feature works well enough and is an improvement overall. I can imagine though that the behaviour when freezed might be unexpected for a user, so disconnecting ADB when freezing / stopping the container may make things clearer. Please let me know what you think. Thanks for your time reviewing!

tools/helpers/net.py
```diff
@@ -0,0 +1,36 @@
...
+import re
+
+def adb_connect(args):
+    """
+    Creates an android debugging connection from the host system to the
+    Waydroid device, if ADB is found on the host system and the device
+    has booted.
+    """
+    # Check if adb exists on the system.
+    if not which("adb"):
```
pymq, line 15: Is my understanding correct that `which` is run under root, but adb is run under user? And so this could be different adb or for instance user have some adb in their PATH, but root haven't?

aleasto: No

aleasto:
Thanks!

PR #887, opened 2023-05-05
fix appmenu folder groups for waydroid

mt190502:
image

### DIFF
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -14,6 +14,40 @@ def start(args, session, unlocked_cb=None):
     waydroid_data = session["waydroid_data"]
     apps_dir = session["xdg_data_home"] + "/applications/"
 
+    def makeMenuFiles():
+        home_directory = os.environ['HOME']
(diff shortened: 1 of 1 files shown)


YukisawaNya:
> not everyone will want this

Almost every Linux developer put their desktop entry under certian group. Only waydroid don't. Why someone in waydroid dev team dislike it and reject it?

Quackdoc:
not on the waydroid team, just an (lesser so now) active community member.

I don't like the idea of forcibly creating a new menu folder. I wouldn't like it in other apps, and I wouldn't like it in waydroid.

adding a command to create the menu should work fine, since it would be a one command/button to create the file. and then it wouldn't need to run again.

aleasto:
There's no reason to have these created at runtime, right?
They could go in the /usr filesystem at packaging time

Quackdoc:
how about a dedicated command for creating the folder? ill make a seperate issue for this but one thing I was thinking of would be a waydroid util function for running various commands/scriptlets, to help converge/fetch popular 3rd scripts. an example of how this would work would be

`waydroid util add-appmenu-folder`

PR #906, opened 2023-05-14
Switch AppArmor profiles into "enforce" mode with changes required by Google Play Services to work

ChrysoliteAzalea:
Hello everyone!

During the last year, [a commit has been merged](https://github.com/waydroid/waydroid/pull/506) that enables AppArmor for the container. In order to avoid breaking any existing installations and work out policy mistakes, all profiles have been switched into the complain mode, in which AppArmor handles domain transitions, logs violations, but doesn't enforce rules (except for deny rules). If proposed AppArmor policies are somehow too restrictive for a container to work properly, it will show up in the audit log.

If you have AppArmor enabled on your system, you can help with AppArmor support in Waydroid:

* by reporting any policy violations by Android container processes in the audit log on your system (/var/log/audit/audit.log if auditd is installed and system log if it's not). They are reported in the AVC category, with "ALLOWED" label if the profile is in complain mode and "DENIED" label if it's in enforce mode
* by switching the profiles into enforce mode and reporting any problems in Waydroid container resulted from that

During [the latest attempt to switch them info *enforce* mode](https://github.com/waydroid/waydroid/pull/725), a problem has been detected that prevented GAPPS images from working. In order to fix that, additional allow rule has been added. I've tried, the official GAPPS image boots fine with that change.

### DIFF
--- a/data/configs/apparmor_profiles/adbd
+++ b/data/configs/apparmor_profiles/adbd
@@ -1,4 +1,4 @@
-profile adbd flags=(attach_disconnected,mediate_deleted,complain) {
+profile adbd flags=(attach_disconnected,mediate_deleted) {
   /** ix,
   /dev** rw,
   network,
--- a/data/configs/apparmor_profiles/android_app
+++ b/data/configs/apparmor_profiles/android_app
@@ -1,4 +1,4 @@
-profile android_app flags=(attach_disconnected, complain, mediate_deleted) {
+profile android_app flags=(attach_disconnected, mediate_deleted) {
   /** ix,
   /dev** rw,
   network,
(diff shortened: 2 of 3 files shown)


aleasto:
> In order to fix that, additional allow rule has been added

Where is it?

ChrysoliteAzalea:
> > In order to fix that, additional allow rule has been added
> 
> Where is it?

Oops, sorry, I've accidentally used ```git commit --gpg-sign --amend``` instead of ```git commit --gpg-sign --amend -a```. Force-pushed with the allow rule added, thank you for noticing.

aleasto:
Did someone properly verify that this doesn't break gapps, magisk, the native bridges, or whatever popular addons people use? Other than "boots fine"

ChrysoliteAzalea:
I'm currently working on issues that can arise from AppArmor (for example, recently I have to add some more permissions to the ```android_app``` profile in order for the LSPosed to work). The main problem here, is that lots of privileged things start from the zygote, and AppArmor permissions are the same for ordinary apps and for system components starting from zygote. On smartphones, zygote handles this by transitioning to a necessary SELinux context, so one of the way to solve it may be patch the zygote in order to add AppArmor support.

I think I'll currently mark it as draft, until it's confirmed that everything works fine in ```enforce``` mode.

PR #944, opened 2023-06-07
Category for the menu placement.

Artiom-M:


Quackdoc:
possible somewhat related: https://github.com/waydroid/waydroid/pull/887

I agree with the system tag for this specifically however

Issue #979, opened 2023-06-26
Is Rooted ADB safe in a privileged container?


ChrysoliteAzalea:
**Is your feature request related to a problem? Please describe.**
Hello everyone! As far as I looked, Waydroid allows an unprivileged user to start a privileged Android container (and, as far as I understand, it cannot be run as an unprivileged container) by making a request to the Waydroid daemon, and then, through Developer settings, enable the "Rooted debugging" option, which allows the user to run ```adb root``` and get root access to the container. Since the container is privileged, the shell would run as an actual root and have system-wide capabilities. The impact is lessened, since the root shell is confined to the container, and its rights are limited by [restriction of capabilities](https://github.com/waydroid/waydroid/pull/504), [container-wide seccomp filter](https://github.com/waydroid/waydroid/pull/505) and [the AppArmor profile](https://github.com/waydroid/waydroid/pull/506), which, however, has not been [switched into enforce mode](https://github.com/waydroid/waydroid/pull/906)

**General information (please complete the following information):**
 - Waydroid tools Version 1.3.3
 - Waydroid Images Version [e.g. VANILLA 20211021], or `system_datetime`/`vendor_datetime` from `/var/lib/waydroid/waydroid.cfg`
 system_datetime = 1682127058
vendor_datetime = 1682162466

**Screenshots:**
![rooted_debugging](https://github.com/waydroid/waydroid/assets/64576901/32cca5ee-4a2c-4459-835c-344e1ec21763)
![adb_run_as_root](https://github.com/waydroid/waydroid/assets/64576901/75290cb7-7e9a-4afd-9e0a-c37d262981ed)

**Desktop (please complete the following information):**
 - OS: Linux Mint 21
 - GPU: Intel Corporation GeminiLake [UHD Graphics 605]
 - Kernel version: 5.15, with Ubuntu patches
 - Host mesa version: 22.2.5-0ubuntu0.1~22.04.3
 - Desktop: Xfce as main, Weston to run Waydroid

**Describe the solution you'd like**
We can write a stricter AppArmor profile for ADBD, to prevent abuse.

**Describe alternatives you've considered**
We can also restrict running ADBD as root, because the root shell can also be safely accessed with ```waydroid shell``` (which is allowed for desktop root only, without creating the risk of privilege escalation).

**Additional context**

Quackdoc:
[...]

PR #989, opened 2023-07-01
Limit ADBD permissions through more restrictive AppArmor profile in order to prevent abuse and privilege escalation by unprivileged host users

ChrysoliteAzalea:
Hello everyone! Last year [I have submitted AppArmor profiles for the Android container](https://github.com/waydroid/waydroid/pull/506). However, these profiles were written in order to prevent abuse by apps. ADBD was not considered dangerous as it is a tool for external intervention. However, [recently I have discovered](https://github.com/waydroid/waydroid/issues/979) that an unprivileged user on a host system can start Waydroid, then enable root debugging and use the ADB in order to get root (which is an actual root on the host since the container is privileged). It is still restricted by being in a container and additionally by AppArmor profiles and seccomp filtering, and it can only get capabilities that are [accessible in the container](https://github.com/waydroid/waydroid/pull/504). However, such restrictions do not seem enough for me, so I propose a more restrictive AppArmor profile for ADBD. In that profile, following restrictions have been added:

* ```/proc``` is now read-only, with write permissions for owner only
* Removing access to abstract Unix sockets (I do not know whether Android uses them)
* ```/var``` and ```/run``` are now read-only, since they seem to be bind-mounted from the host and can be used to manipulate host system
* Some directories I am not sure about are now read-only
* Capabilities are seriously limited in order to ensure nothing system-wide or unsafe for the host can happen in a debugging session
* Mounting and unmounting are disallowed because it can be used to manipulate privileged processes in the container by bind-mounting
* Tracing is restricted because in can be used to manipulate privileged processes
* Signalling is also restricted -- ADBD can only signal itself and apps

These things are confirmed to break:
* ```adb remount``` no longer works, however, it can be done through ```waydroid shell```
* ```adb root``` works, however, its abilities are seriously limited. In order to perform privileged things, users should use ```waydroid shell```, which is only accessible to desktop root


Quackdoc:
does this conflict with #906? if not ill try and test these tommorow and verify I cant run into an issues

ChrysoliteAzalea:
As far as I understand, it shouldn't -- both pull requests switch the **adbd** profile into **enforce** mode.

aleasto:
I'm not sure. Perhaps the feature should be disabled entirely, as apparmor would cover only a subset of waydroid usage.

Quackdoc:
the question is, is there any real world usecase for leaving it enabled?

I don't have strong feelings one way or the other, but I am personally of the opinion that running waydroid (or really any root container) without appropriate SELinux/Apparmor and other security configs stuff should be considered unsafe anyways. I think we should do the best to secure waydroid as much as possible. so if we can do that via Apparmor/SELinux configs without removing *useful* features im all for that. 

that being said if there is no usecase for leaving it enabled, I would prefer outright disabling it

aleasto:
The feature here is being nerfed to dubious usefulness in any case, as per Azalia's own comments

Quackdoc:
profiles are something a user could override if need be, but I genuinely can't think of any use, so I'm for removing it entirely

ChrysoliteAzalea:
I think that we can allow running ADBD as ```shell``` user, but disallow the ```adb root``` since we have other ways to do privileged things (from ```waydroid shell``` to Magisk).

ChrysoliteAzalea:
[I've propose a patch to the ADBD in order to unconditionally block the ```adb root``` command](https://github.com/waydroid/android_vendor_waydroid/pull/22).

ChrysoliteAzalea:
Closing this pull request as [ADBD in Waydroid no longer allows restarting as root](https://github.com/waydroid/android_vendor_waydroid/pull/23).

PR #996, opened 2023-07-09
Allow running a shell command as an arbitrary user and group and with an arbitrary security context

ChrysoliteAzalea:
# Hello everyone!

In this pull request, I propose a new subcommand, ```waydroid custom-shell```, that runs a shell command in the Android container as an arbitrary user, not just as root. It works in a way similar to ```waydroid shell```, by running ```lxc-launch```, but passes more command-line arguments to it. Currently, we have the ```waydroid shell``` subcommand that allows running a command inside the Android container as root. However, unlike the vast majority of other Linux distributions, Android doesn't have any utility similar to ```su``` or ```doas```, which makes running a command as an arbitrary user harder (and not all commands have to be run with the most privileged security context. In this PR, I add another subcommand that accepts more command-line arguments that alllow the user to set UID, GID, SELinux or AppArmor context, and also allow running without AppArmor/seccomp protection, with all capabilities and within the same cgroup (the last three are listed as dangerous as they can leak privileges inside the container, however, I don't think that it's going to be an issue since only the host root is allowed to call this command).

# Alternatives to this pull request

* ```su``` from Magisk -- works, but requires to install Magisk (and I don't like having rooting solutions in a privileged container)
* ```su``` from BusyBox -- works, but requires to have the ```/etc/passwd``` file to be filled with valid entries (can be solved through bind-mounting), not that convenient to use
* patched ```su``` inside the container -- a good idea, but requires writing or adapting an existing ```su``` implementation (it can also require linking with Android libraries, but we can just statically link it with desktop Linux libraries)
* call ```lxc-attach``` directly -- works, but not as convenient
* ```nsenter``` -- a very dangerous alternative, because it does not [clear the capability bounding set](https://github.com/waydroid/waydroid/pull/504), does not enable the ["No New Privileges" restriction](https://github.com/waydroid/waydroid/pull/503), does not [load the seccomp filter](https://github.com/waydroid/waydroid/pull/505) and does not transition to a more secure AppArmor profile. As a result, it can leak privileges inside the container.
* ```run-as``` -- not the same, because it runs a command from a context assigned to an installed app (and does not allow running the command as ```system``` or ```shell``` user, for example), and enforces its own nasty restrictions (that require [some dirty tricks](https://github.com/waydroid/waydroid/pull/994) in order to get it working on non-debuggable packages without patching Android)

### DIFF
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -465,24 +465,24 @@ def custom_shell(args):
         command.append("--gid="+str(args.uid))
     if args.nolsm or args.allcaps or args.nocgroup:
         elevatedprivs = "--elevated-privileges="
-        addpipe = 0
+        addpipe = False
         if args.nolsm:
             if addpipe:
                 elevatedprivs+="|"
             elevatedprivs+="LSM"
-            addpipe = 1
+            addpipe = True
         if args.allcaps:
             if (addpipe == 1):
                 elevatedprivs+="|CAP"
             else:
                 elevatedprivs+="CAP"
-            addpipe = 1
+            addpipe = True
         if args.nocgroup:
             if (addpipe == 1):
                 elevatedprivs+="|CGROUP"
             else:
                 elevatedprivs+="CGROUP"
-            addpipe = 1
+            addpipe = True
         command.append(elevatedprivs)
     if args.context!=None and not args.nolsm:
         command.append("--context="+args.context)
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -12,7 +12,6 @@ import gbinder
 import tools.config
 import tools.helpers.run
 
-
 def get_lxc_version(args):
     if shutil.which("lxc-info") is not None:
         command = ["lxc-info", "--version"]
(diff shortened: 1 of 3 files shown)


aleasto:
I quite prefer the way you implemented this with the first commit only, so as arguments to `waydroid shell`

What was the reason to create `waydroid custom-shell`?

tools/helpers/lxc.py
```diff
@@ -447,6 +446,55 @@ def shell(args):
...
+    elif args.uid!=None:
+        command.append("--gid="+str(args.uid))
+    if args.nolsm or args.allcaps or args.nocgroup:
+        elevatedprivs = "--elevated-privileges="
+        addpipe = 0
+        if args.nolsm:
+            if (addpipe == 1):
+                elevatedprivs+="|LSM"
+            else:
+                elevatedprivs+="LSM"
```
aleasto, line 473:
```suggestion
            if addpipe:
                elevatedprivs+="|"
            elevatedprivs+="LSM"
```

Also please use False/True instead of 0/1 for addpipe

aleasto, line 473:
```suggestion
            if addpipe:
                elevatedprivs+="|"
            elevatedprivs+="LSM"
```

Also please use False/True instead of 0/1 for `addpipe`

ChrysoliteAzalea:
> I quite prefer the way you implemented this with the first commit only, so as arguments to `waydroid shell`
> 
> What was the reason to create `waydroid custom-shell`?

As far as I understand, this PR can change the behavior of ```waydroid shell``` command in an unexpected way, for example, if some script attempts to call it without the ```--``` sequence between the command itself and the command it's going to run, so I decided to leave the original subcommand intact and make another. But as for my system, I'm fine with implementing it as arguments for ```waydroid shell``` subcommand.

tools/helpers/lxc.py
ChrysoliteAzalea: OK, done.

aleasto:
It won't change the behavior. Please do revert back to arguments to `waydroid shell`

tools/helpers/lxc.py
aleasto: Please apply it to all three if cases 😅

ChrysoliteAzalea: Oops, sorry. Done.

aleasto:
Thanks!

ChrysoliteAzalea:
Thank you!

Issue #891, opened 2023-05-07
Fedora Silverblue: urlopen error [Errno 13] Permission denied


plata:
**Describe the bug**
As described in https://www.reddit.com/r/Fedora/comments/136lgp8/waydroid_on_fedora_silverblue/, Waydroid should work on Fedora Silverblue. However, the initial download via GUI fails.

When trying manually, `waydroid init` fails with "Failed to get system OTA channel". Apparently, it doesn't use the default as described in `--help`. Using `-c` and `-v` solves this issue.

**General information:**
 - Waydroid 1.4.1

**Desktop:**
 - OS: Fedora Silverblue (Kinoite) 38

cniw:
on Fedora, user need to write the system and vendor ota url explicitly & manually
- VANILLA image
`sudo waydroid init -c https://ota.waydro.id/system -v https://ota.waydro.id/vendor` (default)
`sudo waydroid init -c https://ota.waydro.id/system -v https://ota.waydro.id/vendor -s VANILLA`
- GAPPS image
`sudo waydroid init -c https://ota.waydro.id/system -v https://ota.waydro.id/vendor -s GAPPS`

plata:
Just out of curiosity: Why is this the case only on Fedora?

The `--help`  should mention it.

cniw:
> Just out of curiosity: Why is this the case only on Fedora?

afaik because of fedora legal/license rules
maybe @aleasto can explain it better since he is the waydroid fedora package maintainer

> The `--help` should mention it.

yea, i think we should mention it

plata:
Ok, anyways even if I do the init manually, it doesn't start:
```
Failed to add service waydroidusermonitor
Failed to add service waydroidclipboard
```
and CPU 100% but nothing happens.

cniw:
> Failed to add service waydroidusermonitor
> Failed to add service waydroidclipboard

can you check the logs with `waydroid log` and `sudo waydroid logcat` or maybe restart your pc

plata:
It shows a lot of
```
Failed to add presence handler: None
```

cniw:
> `Failed to add presence handler: None`

still have no clue :thinking: , do you see the other from this?

plata:
There's
```
conf.c run_buffer: 321 Script exited with status 126
start.c: lxc_end: 988 Failed to run lxc.hook.post-stop for container "waydroid"
```

fsobolev:
[...]

PR #1033, opened 2023-08-01
Clarify 'waydroid init' options ('-c', '-v') on Fedora

plata:
see https://github.com/waydroid/waydroid/issues/891#issuecomment-1537444388


Quackdoc:
Im not sure this should be an error within the script, or if it is, maybe a runtime error that checks distro would be better

plata:
@aleasto can you clarify?

aleasto:
This does not make sense here because Fedora disables it with a downstream patch.
If anything this should be also a downstream patch in Fedora.

Some generic runtime check could also work.

plata:
Are you maintaining that patch? If so, could you remove "; default is Official OTA server" with the patch?

For my curiosity: What's the reason to exclude the default on Fedora?

Issue #903, opened 2023-05-12
RuntimeError: Command failed: % /usr/lib/waydroid/data/scripts/waydroid-net.sh start


yichuang-zh:
**Describe the bug**
After run command "waydroid session start", the terminal shows:
  `[19:24:13] RuntimeError: Command failed: % /usr/lib/waydroid/data/scripts/waydroid-net.sh start
`

Here is the log:
`
(001755) [09:31:46] % modprobe binder_linux devices="anbox-binder,anbox-vndbinder,anbox-hwbinder"
(001755) [09:31:46] Save config: /var/lib/waydroid/waydroid.cfg
(001755) [09:31:48] % mkdir -p /var/lib/waydroid/cache_http
(001755) [09:31:48] Downloading https://sourceforge.net/projects/waydroid/files/images/system/lineage/waydroid_x86_64/lineage-18.1-20230422-GAPPS-waydroid_x86_64-system.zip/download
(001755) [11:07:20] Validating system image
(001755) [11:07:21] Extracting to /var/lib/waydroid/images
(001755) [11:07:27] Save config: /var/lib/waydroid/waydroid.cfg
(001755) [11:07:31] Downloading https://sourceforge.net/projects/waydroid/files/images/vendor/waydroid_x86_64/lineage-18.1-20230422-MAINLINE-waydroid_x86_64-vendor.zip/download
(001755) [11:25:04] Validating vendor image
(001755) [11:25:04] Extracting to /var/lib/waydroid/images
(001755) [11:25:06] Save config: /var/lib/waydroid/waydroid.cfg
(001755) [11:25:06] % modprobe -q ashmem_linux
(001755) [11:25:06] % lxc-info --version
5.0.2
(001755) [11:25:06] % mkdir -p /var/lib/waydroid/lxc/waydroid
(001755) [11:25:06] % sh -c cat "/usr/lib/waydroid/data/configs/config_base" "/usr/lib/waydroid/data/configs/config_3" "/usr/lib/waydroid/data/configs/config_4" > "/var/lib/waydroid/lxc/waydroid/config"
(001755) [11:25:06] % sed -i s/LXCARCH/x86_64/ /var/lib/waydroid/lxc/waydroid/config
(001755) [11:25:06] % cp -fpr /usr/lib/waydroid/data/configs/waydroid.seccomp /var/lib/waydroid/lxc/waydroid/waydroid.seccomp
(001755) [11:25:06] % aa-enabled --quiet
[...]

PR #1062, opened 2023-08-23
Add iptables as Debian dependency

DavidHenryThoreau:


TriMoon:
https://github.com/waydroid/waydroid/issues/903#issuecomment-1658322659

> Wow, iptables has been deprecated from official Debian releases for 4 years, and Waydroid is still experiencing this issue until now.

Quackdoc:
> So instead of this PR, a better solution would be to eliminate the `iptables` usage in the first place...
> Unless it's acceptable to have a follow-up PR that undoes this PR also while trying to get rid of `iptables` in the future..

iptables isn't a requirement, it's a recommendation and the sane default config. distros don't want to ship iptables can ship nftables instead via a simple patch in waydroid-net.sh. nftables has caused issues in the past. 

as for this PR itself, I can't comment on how debian folks handle things, but IMO this should be listed as optional

```diff
-LXC_USE_NFT="false"
+LXC_USE_NFT="true"
```

aleasto:
Ideally we would just figure out how to coexist with docker when using nftables and enable it by default.

Other distributions are already shipping waydroid with ntables, although the status quo in Fedora for example is to use podman, so it's less of an issue

aleasto:
> Wow, iptables has been deprecated from official Debian releases for 4 years, and Waydroid is still experiencing this issue until now.

FYI, `iptables` on modern systems is an alias for `iptables-nft` which is an iptables-compatible frontend for nft.

PR #1075, opened 2023-08-29
container_manager: Keep schedtune cgroup around in case nesting is supported

fredldotme:
Certain devices with Ubuntu Touch make use of the schedtune cgroup boosting mechanism, allowing for better performance system-wide with very little tweaking. Some devices have trouble using this cgroup in a regular GNU/Linux setting because their kernels are not carrying the necessary kernel patch for enabling nesting [here].

To please both those that want Waydroid to stop dropping their performance as well as those having unpatched kernels, probe the capability to nest schedtune cgroups before deciding to unmount it.

[here]: https://gitlab.com/ubports/porting/reference-device-ports/android9/google-pixel-3a/android_kernel_google_bonito/-/commit/6452429d1f1dbdc9e1e310e82cc18664f2d1b56d

### DIFF
--- a/tools/actions/container_manager.py
+++ b/tools/actions/container_manager.py
@@ -146,9 +146,20 @@ def do_start(args, session):
     if which("start"):
         command = ["start", "cgroup-lite"]
         tools.helpers.run.user(args, command, check=False)
+
+    # Keep schedtune around in case nesting is supported
     if os.path.ismount("/sys/fs/cgroup/schedtune"):
-        command = ["umount", "-l", "/sys/fs/cgroup/schedtune"]
-        tools.helpers.run.user(args, command, check=False)
+        try:
+            os.mkdir("/sys/fs/cgroup/schedtune/probe0")
+            os.mkdir("/sys/fs/cgroup/schedtune/probe0/probe1")
+        except:
+            command = ["umount", "-l", "/sys/fs/cgroup/schedtune"]
+            tools.helpers.run.user(args, command, check=False)
+        finally:
+            if os.path.exists("/sys/fs/cgroup/schedtune/probe0/probe1"):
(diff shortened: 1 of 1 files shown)


aleasto (approved):

tools/actions/container_manager.py
```diff
@@ -146,9 +146,20 @@ def do_start(args, session):
...
         command = ["start", "cgroup-lite"]
         tools.helpers.run.user(args, command, check=False)
+
+    # Keep schedtune around in case nesting is supported
     if os.path.ismount("/sys/fs/cgroup/schedtune"):
-        command = ["umount", "-l", "/sys/fs/cgroup/schedtune"]
-        tools.helpers.run.user(args, command, check=False)
+        try:
+            os.mkdir("/sys/fs/cgroup/schedtune/probe0")
+            os.mkdir("/sys/fs/cgroup/schedtune/probe0/probe1")
```
peat-psuwit, line 154: Won't this call throw on unpatched kernel? In that case, it's probably better to move `umount` to the `except:` block. Also moving `os.rmdir()` call to `finally:` block might reduce too-broad try-except block.

fredldotme: Addressed

fredldotme: Note that I'm not much of a fan of relying on exceptions in C++, but I guess it's fine for Python.

peat-psuwit: It's actually pretty common for Python code to rely for exception for normal error handling. This concept is called "Easier to ask for forgiveness than permission" [1]. Besides, `os.mkdir()` will throw whether or not you like it.

[1] https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

peat-psuwit (approved):

PR #1083, opened 2023-09-03
Add an option to decide whether to create a desktop file

OneRain233:
Add new options `show_waydroid_apps` in `waydroid.cfg` to decide whether to create a dot desktop file for all apps.

### DIFF
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -22,7 +22,8 @@ config_keys = ["arch",
                "system_datetime",
                "vendor_datetime",
                "suspend_action",
-               "mount_overlays"]
+               "mount_overlays",
+               "show_waydroid_apps"]
 
 # Config file/commandline default values
 # $WORK gets replaced with the actual value for args.work (which may be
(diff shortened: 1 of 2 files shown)


Quackdoc:
It would be nice to add docs for this... somewhere

aleasto:
We *must* create the desktop files otherwise the windows won't have an icon and will display the Wayland icon instead

aleasto:
How does wine handle this on Wayland?
I know you can disable the wine menubuilder, but surely they must be installing desktop files somewhere anyways so that they can give icons to the windows...

Speaking of, i'd like this config to be called menu_builder as well, i think it makes sense. If we can land it...

OneRain233:
You're right, I didn't consider the specifics around Wine, this option may not be entirely robust yet.

Quackdoc:
Instead of not creating them, perhaps it would be better to add `NoDisplay=true`? I suspect that is how wine does it

Twig6943:
> Instead of not creating them, perhaps it would be better to add `NoDisplay=true`? I suspect that is how wine does it

Seems to be the case

https://gitlab.winehq.org/search?group_id=10&nav_source=navbar&project_id=5&repository_ref=master&search=NoDisplay&search_code=true

@OneRain233 would you be down to finish this pr or do I just pick up where you left off?

Twig6943:
Yo why was this closed?

PR #1091, opened 2023-09-13
debian: depend on iptables

JamiKettunen:
`USE_NFTABLES=0` is the default the package is built with, yet it doesn't depend on `iptables` which may be missing and breaks launching Waydroid due to networking script failing with `IP*TABLES_BIN` variables ending up empty.


aleasto:
Duplicate of #1062

PR #1096, opened 2023-09-17
Enable the "ro.adb.secure" property

ChrysoliteAzalea:
Hello everyone! I propose enabling the ```ro.adb.secure``` property.

**Explanation**:

Android Debug Bridge is a tools that allows remote communication with the Android system. On the Android system side, it's represented by the Android Debug Bridge Daemon, that runs as a special user (```shell```). ```ro.adb.secure``` is a property that allows ADBD to be managed by Android settings. If ```ro.adb.secure``` is set to ```0```, anyone that can connect to the Android system over network (including installed Android apps connecting to ```127.0.0.1```) can get access to the privileged shell with lots of irrevocable Android permissions, and remotely control the system through tools like ```scrcpy```. If it's set to ```1```, all such interactions have to be authorized by the user -- it cannot be done over network.

**Possible benefits**:

* ADB can be disabled in the settings (without re-enabling on container restart)
* ADB connections have to be authorized by user
* Such authorizations can be revoked

**Possible problems**:

I haven't found any.

### DIFF
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -244,6 +244,10 @@ def make_base_props(args):
     if not os.path.exists("/dev/ashmem"):
         props.append("sys.use_memfd=true")
 
+# Added for security reasons
+    props.append("ro.adb.secure=1")
+
+
     egl = tools.helpers.props.host_get(args, "ro.hardware.egl")
     dri, _ = tools.helpers.gpu.getDriNode(args)
 
@@ -260,6 +264,7 @@ def make_base_props(args):
             egl = "swiftshader"
         props.append("debug.stagefright.ccodec=0")
     props.append("ro.hardware.gralloc=" + gralloc)
+
 
     if egl != "":
         props.append("ro.hardware.egl=" + egl)
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -244,11 +244,9 @@ def make_base_props(args):
     if not os.path.exists("/dev/ashmem"):
         props.append("sys.use_memfd=true")
 
-# Added for security reasons
+    # Added for security reasons
     props.append("ro.adb.secure=1")
     props.append("ro.debuggable=0")
-
-
     egl = tools.helpers.props.host_get(args, "ro.hardware.egl")
     dri, _ = tools.helpers.gpu.getDriNode(args)
 
@@ -265,7 +263,6 @@ def make_base_props(args):
             egl = "swiftshader"
         props.append("debug.stagefright.ccodec=0")
     props.append("ro.hardware.gralloc=" + gralloc)
-
 
     if egl != "":
         props.append("ro.hardware.egl=" + egl)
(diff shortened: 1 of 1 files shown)


Quackdoc:
it's worth noting that this may fix some apps (not all, but some) that block waydroid, I do know at least some apps for check this. I think it's a good idea, users should always be able to revert this anyways in prop file

Entropy512:
It's been a looooooooooong time since I've done Android platform development, but the primary use case (e.g. the reason I did it many times back in the day) for this would be during platform bringup in order to support debugging in a scenario where you can't even get to Settings without something crashing.

But platform developers (in this case that would be Waydroid developers) can easily change this if needed, and I can't think of any legitimate reason for setting adb.secure to 0 on an end-user-facing release build.

ChrysoliteAzalea:
> It's been a looooooooooong time since I've done Android platform development, but the primary use case (e.g. the reason I did it many times back in the day) for this would be during platform bringup in order to support debugging in a scenario where you can't even get to Settings without something crashing.

I understand why it may be necessary during development, but it's just not safe leaving it that way. Also, we have the root shell from ```waydroid shell```. ADB has lots of rights on the platform (although it's not root), such as seeing external storage, installing apps, setting them as device owners, etc, so it's not safe to allow ADB shell over the Internet.

JamiKettunen:
@aleasto Thoughts?

JamiKettunen:
All I really have to say is not sure what's up with the extra newline additions and non-indented comment

ChrysoliteAzalea:
> All I really have to say is not sure what's up with the extra newline additions and non-indented comment

Seems to be accidental.

ChrysoliteAzalea:
OK, fixed.

tools/helpers/lxc.py
```diff
@@ -244,6 +244,9 @@ def find_hidl(intf):
     if not os.path.exists("/dev/ashmem"):
         props.append("sys.use_memfd=true")
 
+    # Added for security reasons
+    props.append("ro.adb.secure=1")
+    props.append("ro.debuggable=0")
```
JamiKettunen, line 249: Hm actually above replacing `props = []` probably is better since they're always there anyway, no need to unconditionally append to the end later:
```py
    props = [
        "ro.adb.secure=1",
        "ro.debuggable=0",
    ]
```
Either way would be nice to hear if this would even be considered for merging ^^

aleasto:
I'm torn between having it here or letting it be decided by the system image while switching the defaults on the official one.

JamiKettunen:
@aleasto Decided yet? One of my apps seemingly has now started using something called "RootBeer" (which I could see in `waydroid logcat`) and specifically it fails on exactly these props https://github.com/scottyab/rootbeer/blob/master/rootbeerlib/src/main/java/com/scottyab/rootbeer/RootBeer.java#L258-L290

It seems they already blocked me on server side though or do even more additional checks that I'm yet to find out about...

Joabvoss1977 (approved):

Fuseteam:
@aleasto still torn how to proceed?

PR #1103, opened 2023-09-24
user_manager: Respect user umask and use less code

Arusekk:
chmod(0o644) is no longer necessary after umask(0) removal in da4772c4e54467920d642e2a792c5d16d3b7bf33 ("tools: Remove umask 0") as the default umask is 0o022, but if the umask is stricter (e.g. 0o027), then the user would expect the created files to be 0o640. Creating the directory with the correct mode already also avoids potential race conditions on systems with insufficient umask. ~~Also, by reversing the logic to the intended flow, it avoids issues mentioned in #817.~~

While at it, also use templating strings for .desktop file creation to simplify the logic and split the template from the actual code a bit.

### DIFF
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -29,40 +29,38 @@ def start(args, session, unlocked_cb=None):
 
         desktop_file_path = apps_dir + "/waydroid." + packageName + ".desktop"
         if not os.path.exists(desktop_file_path):
-            lines = ["[Desktop Entry]", "Type=Application"]
-            lines.append("Name=" + appInfo["name"])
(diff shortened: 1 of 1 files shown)


aleasto:
> by reversing the logic to the intended flow

I'm sorry, what do you mean by that?

aleasto:
Note that we have to keep the `os.remove` before overwriting the file otherwise some desktops (mine, KDE Plasma) may not update their database. Presumably they have an `inotify` setup to only monitor file addition/deletion

aleasto:
The other fixes are good so i'll merge this version for now

PR #1149, opened 2023-10-31
Add an XDG menu directory

lcorbasson:
As of today, Waydroid apps are added to the XDG apps menu without a dedicated menu directory. This PR aims to fix that, to ensure that:
- Waydroid apps are clearly identified and cannot be mistaken for their desktop counterparts (e.g. Firefox/Android and Firefox/desktop),
- Waydroid apps don't end up in a "Lost & Found" menu directory generated by the user's desktop environment.

This approach is coherent with what other emulation layers do – e.g. WINE or Chrome Apps.

It is based on @mt190502's work in https://github.com/waydroid/waydroid/pull/887, but with the idea that static menu/directory files should not be (re-)generated at runtime but shipped with the software (package). This also allows the user to remove/deactivate them more easily, including using his desktop environment's tools for this purpose.

The end-user result will be as following (adjusted for the user's XDG-compliant desktop environment):
> image


aleasto:
Yes i definitely agree with this.

Do you know if the user can disable the grouping, for example by overriding the desktop file in `~/.config/menu`?

lcorbasson:
Yes. The user simply needs to add an empty file as `~/.config/menus/applications-merged/waydroid.menu` and it will override the system-wide `waydroid.menu`.

lcorbasson:
Thanks!

lcorbasson:
@aleasto You may want to close https://github.com/waydroid/waydroid/pull/887 too.

firefoxlover:
this is great! But the .desktop entries will still be placed in the main ~/.local/share/applications directory, right?

lcorbasson:
Yes, it doesn't change the code for the Waydroid apps' .desktop files. It only creates a menu entry for all .desktop files with a X-WayDroid-App category.

PR #1150, opened 2023-11-01
Don't change user-defined desktop files for Waydroid when they're set as read-only

lcorbasson:
This PR is a proposal to solve PR https://github.com/waydroid/waydroid/pull/817.

- On usual installations, it won't change a thing to current behaviour.
- On installations where a user may want to preserve their changes to the `Waydroid.desktop` file, they now can by setting it as read-only (`chmod -r`, Windows readonly attribute, etc.).


aleasto:
Do you see a reason for removing the file before re-creating it instead of overwriting it?

PR #1173, opened 2023-11-19
update regex for deprecation warning

aarron-lee:
Python interprets \d as invalid escape sequence.  In a future Python version there will be a SyntaxWarning and eventually a SyntaxError.

Fix this via declaring the RegEx pattern as a raw string instead by prepending r

image


aleasto:
Thanks

Issue #824, opened 2023-03-22
Waydroid subcommands disagree whether waydroid is initialized


flash-freezing-lava:
**Describe the bug**
`waydroid session init` claims the session is not initialized, but according to `waydroid init` it is.
```
$ waydroid session init
ERROR: WayDroid is not initialized, run "waydroid init"
# waydroid init
[08:29:55] Already initialized
```

**General information:**
 - Waydroid tools Version 1.4.0
 - Waydroid Images Version `system_datetime=1679116647`/`vendor_datetime=1679138482` from `/var/lib/waydroid/waydroid.cfg`

**Desktop:**
 - OS: ArchLinux x64
 - GPU: NVIDIA GTX1070
 - Kernel version: 6.2.7-arch1-1
 - Host mesa version: 23.0.0-3
 - Desktop: gnome-desktop 43.2-1

**Logs (please upload as file)**
 - Prop file [waydroid_base.prop.txt](https://github.com/waydroid/waydroid/files/11037497/waydroid_base.prop.txt) at `/var/lib/waydroid/waydroid_base.prop`
 - Log file [waydroid.log](https://github.com/waydroid/waydroid/files/11037509/waydroid.log) at `/var/lib/waydroid/waydroid.log`
 - Output of `waydroid logcat` is also `ERROR: WayDroid is not initialized, run "waydroid init"`
 - Output of `dmesg` [dmesg.log](https://github.com/waydroid/waydroid/files/11037536/dmesg.log)

cniw:
> waydroid session init

wrong command, `init` not availale for `session`, it should be `start`
```bash
waydroid session start
```

flash-freezing-lava:
I mistyped it in the issue. The error occurred with `waydroid session start`.

inemum:
I was having the same problem because `/var/lib/waydroid` had the wrong permissions (it was only accessible to root).

flash-freezing-lava:
Thanks for pointing this out.

Changing permissions of `/var/lib/waydroid` and `/dev/anbox-binder` solved my issues.

To help users find their issue easier in the future, a warning in the main function could help them. As a simple version, I'm thinking of something like:
```python
mode = os.stat(args.work).st_mode
if mode & stat.S_IROTH == 0:
  print("WARNING: The globally created waydroid files are unreadable by regular users. "
    "Please ensure manually that all files in %s are world-readable. "
    "This might also apply to binderfs devices."
    % args.work, file=sys.stderr)
```
[...]

PR #1190, opened 2023-12-10
Ensure default umask to avoid surprises with file permissions.

es-fabricemarie:
Fixes waydroid/waydroid#824


sertonix:
It would be nice to have this merged

aleasto:
I think this is ok

Issue #1131, opened 2023-10-15
[BUG] Clipboard: copy is working, but not pasting


sloonz:
### Describe the bug

Copying text (from android to host) works fine, but paste (from host to android) does nothing.

While pressing "paste", I got this output in logcat : 

```
10-15 14:26:25.362   266   879 I system_server: oneway function results will be dropped but finished with status OK and parcel size 4
10-15 14:26:25.362   266  2621 I system_server: oneway function results will be dropped but finished with status OK and parcel size 4
10-15 14:26:25.362   266  2621 I system_server: oneway function results will be dropped but finished with status OK and parcel size 4
10-15 14:26:25.496    76   195 W APM::AudioPolicyEngine: getDevicesForStrategy() unknown strategy: -1
10-15 14:26:25.496   266  2621 I system_server: oneway function results will be dropped but finished with status OK and parcel size 4
10-15 14:26:25.526  2595  2635 D OpenGLRenderer: endAllActiveAnimators on 0x7f5a13698200 (RippleDrawable) with handle 0x7f5a13613e40
10-15 14:26:28.398   266   879 I system_server: oneway function results will be dropped but finished with status OK and parcel size 4
```

### Waydroid version

1.4.1

### Device

Linux Desktop

### Operating System

ArchLinux

### Kernel version

6.5.7-arch1-1

### Desktop Environment

Sway  1.8.1

### GPU

Radeon RX 7800 XT

### Logs

/var/lib/waydroid/waydroid.cfg

```
[waydroid]
arch = x86_64
vendor_type = MAINLINE
system_datetime = 1697247022
vendor_datetime = 1697282537
suspend_action = freeze
mount_overlays = True
images_path = /var/lib/waydroid/images
system_ota = https://ota.waydro.id/system/lineage/waydroid_x86_64/GAPPS.json
vendor_ota = https://ota.waydro.id/vendor/waydroid_x86_64/MAINLINE.json
binder = binder
vndbinder = anbox-vndbinder
hwbinder = anbox-hwbinder
binder_protocol = aidl3
service_manager_protocol = aidl3

[properties]
```

/var/lib/waydroid/waydroid.prop

```
sys.use_memfd=true
debug.stagefright.ccodec=0
ro.hardware.gralloc=gbm
ro.hardware.egl=mesa
ro.hardware.vulkan=radeon
ro.hardware.camera=v4l2
ro.opengles.version=196609
[...]

PR #1207, opened 2023-12-21
Fix Host to guest paste clipboard issues

meramsey:
This fix worked for me to solve the "Fix Host to guest paste clipboard issues" with my setup. 

It should hopefully solve: https://github.com/waydroid/waydroid/issues/1131

Found after alot of debugging that both pyclip/pyperclip failed to ever get the clipboard data from the paste. 

confirmed via xsel manually clipboard was fine for my DISPLAY which was :99 in my case
```
export DISPLAY=:99; while true; do xsel --output --clipboard; sleep 1; done
```

So added that as a fallback and seems to solve the issue and i can copy/paste both directions now without issues.

### DIFF
--- a/tools/services/clipboard_manager.py
+++ b/tools/services/clipboard_manager.py
@@ -17,14 +17,36 @@ def start(args):
     def sendClipboardData(value):
         try:
             pyclip.copy(value)
+        except UnicodeDecodeError as e:
+            logging.debug("sendClipboardData: UnicodeDecodeError: %s", str(e))
+            logging.debug("During handling of the above exception, the data causing the error was: %s", value)
         except Exception as e:
-            logging.debug(str(e))
+            logging.debug(f"sendClipboardData: Exception: {str(e)} occurred with {value}")
+
+    def get_clipboard_via_xsel():
+        import subprocess
+        try:
+            return subprocess.check_output(['xsel', '--output', '--clipboard']).decode('utf-8')
+        except subprocess.CalledProcessError as e:
+            print("An error occurred while trying to read the clipboard using xsel.")
+            return ""
 
     def getClipboardData():
         try:
+            # Attempt to get clipboard data using pyclip
             return pyclip.paste()
         except Exception as e:
-            logging.debug(str(e))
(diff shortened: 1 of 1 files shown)


aleasto:
This is a Wayland project, we wouldn't want to use the X clipboard.

meramsey:
Thank you for the feedback. I understand the project's emphasis on Wayland. However, the proposed solution using `xsel` is intended as a fallback mechanism specifically for scenarios where the primary method (pyclip) fails. This approach ensures broader compatibility and reliability in clipboard operations, especially in mixed environments where Wayland and X11 coexist. Utilizing `weston --backend=x11-backend.so --xwayland &` demonstrates that X11 support is inherently accommodated within Wayland environments, thus maintaining compatibility without detracting from the Wayland-centric approach of Waydroid.

Further references:
- [Weston with XWayland on Ubuntu Manpages](https://manpages.ubuntu.com/manpages/focal/en/man1/weston.1.html#xwayland)
- [Weston Manpage on Arch Linux](https://man.archlinux.org/man/weston.1.en#XWAYLAND)

meramsey:
@aleasto Any thoughts on this

Quackdoc:
if it's just a fallback it's probably not the worst thing, however xsel is an entirely seperate package anyways that many users may not have, if we are going to rely on a third party package, Perhaps one exists that sufficiently covers both non xwayland and x11?

meramsey:
I haven't found any others that worked as a fallback, but If someone does find one I'd recommend adding them similarly as fallbacks so there are different options it can automatically try to use in those cases.

meramsey:
Been using the fix for a long time now without issues not seen any issues and its only a fallback. So if whoever has the permissions to merge after testing to confirm it works for them it should be a solved issue.

aleasto:
Waydroid now uses the Wayland clipboard directly, without interacting with `pyclip` and `wl-clipboard` anymore. This is officially out of scope at the present moment. If you wish your clipboard to be synced with X when using nested compositors, you should look into a solution that synchronizes the X clipboard with the nested compositor itself.

PR #1216, opened 2023-12-24
Add prefix to all install directories

SuperSandro2000:
Otherwise the installs fail on NixOS for 1.4.2 because the application is not allowed to write there in the build sandbox.


aleasto:
That doesn't make sense 
You override the install destination with DESTDIR, not with PREFIX

SuperSandro2000:
True and it also shouldn't have worked since /usr/etc should be as unwrittable as /etc 🤔. I need to take a second look on this.

SuperSandro2000:
It worked because in nixpkgs PREFIX was set instead of DESTDIR. 🤦🏼

PR #1218, opened 2023-12-25
Add prefix to all install directories

SuperSandro2000:
I got confused in #1216. In NixOS we want to overwrite the PREFIX, so that waydroid thinks it is installed into the nix store path instead of just installing the directories to there.

See https://github.com/NixOS/nixpkgs/pull/276409#discussion_r1435881177

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,13 @@ WAYDROID_DIR := $(PREFIX)/lib/waydroid
 BIN_DIR := $(PREFIX)/bin
 APPS_DIR := $(PREFIX)/share/applications
 APPS_DIRECTORY_DIR := $(PREFIX)/share/desktop-directories
-APPS_MENU_DIR := /etc/xdg/menus/applications-merged
+APPS_MENU_DIR := $(PREFIX)/etc/xdg/menus/applications-merged
 METAINFO_DIR := $(PREFIX)/share/metainfo
 ICONS_DIR := $(PREFIX)/share/icons
 SYSD_DIR := $(PREFIX)/lib/systemd/system
 DBUS_DIR := $(PREFIX)/share/dbus-1
 POLKIT_DIR := $(PREFIX)/share/polkit-1
-APPARMOR_DIR := /etc/apparmor.d
+APPARMOR_DIR := $(PREFIX)/etc/apparmor.d
 
 INSTALL_WAYDROID_DIR := $(DESTDIR)$(WAYDROID_DIR)
 INSTALL_BIN_DIR := $(DESTDIR)$(BIN_DIR)


aleasto:
I don't know what Nix is doing, but `$(PREFIX)/etc/` is not going to work for every other packaging system

SuperSandro2000:
Nix installs the file into the nix store at ``/nix/store/07pqx764krx7467h657rv74calrg98qh-waydroid-1.4.2/etc/`` in my case where they also live in the build system and from there creates symlinks. If the DESTDIR is used then the bin/waydroid symlink points into the void and doesn't work.

```
 ➜ tree result
 result ⇒ /nix/store/3dy5l36gyyaacwrj2wl9h933c38nsqav-waydroid-1.4.2
├──  bin
│   └──  waydroid ⇒ /lib/waydroid/waydroid.py
```

aleasto:
The solution is to make it a relative symlink

SuperSandro2000:
It would be relative to PREFIX. The /nix/store/... directory is not going away at runtime, so PREFIX is correct here, not DESTDIR

aleasto:
It would be relative to the waydroid.py file, which is what you want:
`ln -s ../bin/waydroid $(WAYDROID_DIR)/waydroid.py`

`$(PREFIX)/etc` doesn't work because nobody puts `/etc` under `/usr`.

jtojnar:
Usually, this is handled by using separate `sysconfdir` variable that defaults to `/etc`, see https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Installation-Directory-Variables.html

CajuM:
> It would be relative to the waydroid.py file, which is what you want: `ln -s ../lib/waydroid.py $(INSTALL_BIN_DIR)/waydroid`
> 
> `$(PREFIX)/etc` doesn't work because nobody puts `/etc` under `/usr`.

According to https://www.gnu.org/prep/standards/html_node/Directory-Variables.html `sysconfdir` is normally `$(prefix)/etc` but can be overridden, something like this should work:
```make
SYSCONFDIR := $(PREFIX)/etc
APPS_MENU_DIR := $(SYSCONFDIR)/xdg/menus/applications-merged
```

aleasto:
I'm ok with something like that, but not with defaulting `SYSCONFDIR` to `$(PREFIX)/etc`, it's not a thing in fedora nor debian or arch

jtojnar:
> According to [gnu.org/prep/standards/html_node/Directory-Variables.html](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) `sysconfdir` is normally `$(prefix)/etc` but can be overridden, something like this should work:

Right, sorry, I was thinking [Meson](https://mesonbuild.com/Builtin-options.html#universal-options), which special-cases the default:

> When the prefix is `/usr`: `sysconfdir` defaults to `/etc`

That is probably unnecessarily complex for us, we can default to `/etc` always, the important part is having the path overridable.

CajuM:
Well, these are just defaults, the distro should be able to configure them.

CajuM:
How about?:
```diff
diff --git a/Makefile b/Makefile
index 5c7bdef..d1cbc3a 100644
--- a/Makefile
+++ b/Makefile
@@ -4,17 +4,18 @@ USE_SYSTEMD ?= 1
 USE_DBUS_ACTIVATION ?= 1
 USE_NFTABLES ?= 0
 
+SYSCONFDIR := /etc
 WAYDROID_DIR := $(PREFIX)/lib/waydroid
 BIN_DIR := $(PREFIX)/bin
 APPS_DIR := $(PREFIX)/share/applications
 APPS_DIRECTORY_DIR := $(PREFIX)/share/desktop-directories
-APPS_MENU_DIR := $(PREFIX)/etc/xdg/menus/applications-merged
+APPS_MENU_DIR := $(SYSCONFDIR)/xdg/menus/applications-merged
 METAINFO_DIR := $(PREFIX)/share/metainfo
 ICONS_DIR := $(PREFIX)/share/icons
 SYSD_DIR := $(PREFIX)/lib/systemd/system
 DBUS_DIR := $(PREFIX)/share/dbus-1
 POLKIT_DIR := $(PREFIX)/share/polkit-1
-APPARMOR_DIR := $(PREFIX)/etc/apparmor.d
+APPARMOR_DIR := $(SYSCONFDIR)/apparmor.d
 
 INSTALL_WAYDROID_DIR := $(DESTDIR)$(WAYDROID_DIR)
 INSTALL_BIN_DIR := $(DESTDIR)$(BIN_DIR)
```

aleasto:
Looks good. Can you propose it in a MR instead?

PR #1239, opened 2024-01-16
Resumable downloads during initial setup

bemoody:
When installing waydroid, if the download is interrupted (e.g. because the network died, or you killed the process halfway through, or the server randomly broke the connection), it should be possible to keep going from where you left off, rather than re-downloading a 700+ MB file from the beginning.

This is a quick and dirty implementation to do pretty much what `wget -c` does.

(I'm not sure there's much benefit in waydroid having its own kludgy download function rather than using Wget; but this ought to work.)

### DIFF
--- a/tools/helpers/http.py
+++ b/tools/helpers/http.py
@@ -99,25 +99,45 @@ def download(args, url, prefix, cache=True, loglevel=logging.INFO,
     prefix = prefix.replace("/", "_")
     path = (args.work + "/cache_http/" + prefix + "_" +
             hashlib.sha256(url.encode("utf-8")).hexdigest())
+    start_pos = 0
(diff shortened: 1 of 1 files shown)


aleasto:
If i read this correctly: if the download is corrupted but big enough, trying to download again will instead reuse the cached file and keep failing.

bemoody:
Good point, thanks.

I only see two places where `helpers.http.download` is used (are there any others?)  In both of those cases, *if the SHA-256 hash doesn't match, the file is deleted.*

On the other hand, looking at it now, I notice that I totally messed up the handling of 416 errors.  So, if the file is already fully downloaded, you'd get stuck, not because the checksum is wrong but because the `download` function would keep raising exceptions.

PR #1294, opened 2024-02-21
Adds config support for overlays_readonly

jeremyselan:
Some users workflows rely upon bind mounts for various filesystem paths, including beneath /vendor, and beneath /system. But assuming overlays to be readonly precludes such usage.  This commit exposes a configuration option, "overlays_readonly", for such users. The default behavior is unchanged.

### DIFF
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -23,7 +23,8 @@ config_keys = ["arch",
                "vendor_datetime",
                "suspend_action",
                "mount_overlays",
-               "auto_adb"]
+               "auto_adb",
+               "overlays_readonly"]
 
 # Config file/commandline default values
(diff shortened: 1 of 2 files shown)


aleasto:
I'm not sure I understand the workflow that you're mentioning.

Let me give you some talking points, even though they might not be what you're looking for...

If you know in advance what directories or files you need added, you can put them in `/var/lib/waydroid/overlay/[vendor/]`.

Otherwise, the original idea was that you'd just `mount -o rw,remount` the overlay and start hacking around.
But something changed (kernel side?) and `mount -o rw,remount /var/lib/waydroid/rootfs` now fails in distros like Arch and Fedora. You can still `sudo mount -o rw,remount /` from `sudo waydroid shell` though apparently.

Issue #1310, opened 2024-03-05
[Feature] Add menu shortcut to stop Waydroid by default


raploz:
### Is your feature request related to a problem? Please describe.

Waydroid handily adds menu shortcuts for starting itself and for starting apps, however, if you just close it it keeps running in the background.

### Describe the solution you'd like.

It would be good if it could also add a "Stop Waydroid" shortcut to completely stop it in the background, the same way the command `sudo waydroid session stop` does. This would greatly improve usability.

### Describe alternatives you've considered.

A shortcut is the easiest solution for developers, I imagine. Another, more complex option, would be a system tray icon indicating whether the session is started and with an option to quit. This might prove troublesome for GNOME users though, so I think a simple shortcut should suffice.

lidgnulinux:
Just make a shortcut or desktop file, it's not hard at all !

```
$ touch stop_waydroid.sh
```

```
$ cat >stop_waydroid.sh<<__EOF__
waydroid session stop
pkexec systemctl stop waydroid-container
__EOF__
```

PR #1311, opened 2024-03-05
Add shortcut to quit Waydroid

raploz:
I'm not a developer, so I'm not sure if this is the right thing, but I made #1310  and this is an attempt at fixing it. Hopefully it's all it takes?


dos1:
I'd rather add an action entry in the existing Waydroid .desktop file than create a whole separate launcher.

PR #1315, opened 2024-03-07
Use synthetic paths for host sockets

Joshua-Ashton:
Mount XDG_RUNTIME_DIR, PULSE_RUNTIME_PATH and
WAYLAND_DISPLAY to our own synthetic paths in
/waydroid/xdg/...

That way, they can never clash with any of our
Android bindpoints.

Eg. /data/.../xdg/wayland-0 conflicts with /data

### DIFF
--- a/tools/helpers/images.py
+++ b/tools/helpers/images.py
@@ -136,10 +136,10 @@ def make_prop(args, cfg, full_props_path):
     add_prop("waydroid.host.uid", "user_id")
     add_prop("waydroid.host.gid", "group_id")
     add_prop("waydroid.host_data_path", "waydroid_data")
-    add_prop("waydroid.xdg_runtime_dir", "xdg_runtime_dir")
-    add_prop("waydroid.pulse_runtime_path", "pulse_runtime_path")
-    add_prop("waydroid.wayland_display", "wayland_display")
     add_prop("waydroid.background_start", "background_start")
+    props.append("waydroid.xdg_runtime_dir=" + tools.config.defaults["container_xdg_runtime_dir"])
+    props.append("waydroid.pulse_runtime_path=" + tools.config.defaults["container_pulse_runtime_path"])
+    props.append("waydroid.wayland_display=" + tools.config.defaults["container_wayland_display"])
     if which("waydroid-sensord") is None:
         props.append("waydroid.stub_sensors_hal=1")
     dpi = cfg["lcd_density"]
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -41,6 +41,8 @@ defaults = {
     "suspend_action": "freeze",
     "mount_overlays": "True",
     "auto_adb": "True",
+    "container_xdg_runtime_dir": "/run/xdg",
+    "container_wayland_display": "wayland-0",
 }
 defaults["images_path"] = defaults["work"] + "/images"
 defaults["rootfs"] = defaults["work"] + "/rootfs"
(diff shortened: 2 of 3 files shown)


Quackdoc:
I was only able to check on my phone but I don't see any glaring issues with it, @aleasto I believe would need to merge it, im not sure if erfan or jon are still active

aleasto:
Big +1 for the change, but there's no need to make it a session config

aleasto:
Hi... Sorry it's been so long.

The session{} configs are meant for variables that depend on the host user session (like socket paths, user id, etc...).
This change instead is aimed at the opposite direction, that is making these paths independent from the user session, so I've moved the synthetic path definitions to defaults{} instead.

I've also had to change the root directory from `/waydroid` to `/run`, because it must be under an existing mountpoint:
* https://github.com/waydroid/android_vendor_waydroid/blob/5a57eacfa864bb38b4c7c21630fbc32b13a729d6/BoardConfigExtra.mk#L20
* https://github.com/waydroid/waydroid/blob/3dd05999699513b687b09b1cf8dfbd92063c783d/tools/helpers/lxc.py#L114-L117

Does this still match your needs?

Joshua-Ashton:
It should work for my needs, but `/waydroid` worked for me when I tried it.

Perhaps I had some other config setting that up. We have a lot of custom stuff.

aleasto:
🤷 LXC should fail to mount files under `/waydroid` because `/waydroid` does not exist (in the default OTA images anyways) and `/` is mounted read-only.

Joshua-Ashton:
Ah, we do not mount as read-only, which is probably why

Joshua-Ashton:
wrt mounting as read only: https://github.com/waydroid/waydroid/pull/1294

tools/helpers/images.py
```diff
@@ -136,10 +136,10 @@ def add_prop(key, cfg_key):
     add_prop("waydroid.host.uid", "user_id")
     add_prop("waydroid.host.gid", "group_id")
     add_prop("waydroid.host_data_path", "waydroid_data")
-    add_prop("waydroid.xdg_runtime_dir", "xdg_runtime_dir")
-    add_prop("waydroid.pulse_runtime_path", "pulse_runtime_path")
-    add_prop("waydroid.wayland_display", "wayland_display")
     add_prop("waydroid.background_start", "background_start")
+    props.append("waydroid.xdg_runtime_dir=" + tools.config.defaults["container_xdg_runtime_dir"])
```
linsui, line 140: I don't know why but this change makes waydroid not start on NixOS. I created `/run/xdg` manually but it doesn't help.

aleasto: Are you doing something funny with `/var/lib/waydroid/lxc/waydroid/config_session`?
It will populate the android filesystem with `/run/xdg`

The file is generated whenever starting Waydroid, so it should always be correct unless you're manipulating it in funny ways.
Mine looks like this:
```
lxc.mount.entry = tmpfs /run/xdg none create=dir 0 0
lxc.mount.entry = /run/user/1000/wayland-0 run/xdg/wayland-0 none rbind,create=file 0 0
lxc.mount.entry = /run/user/1000/pulse/native run/xdg/pulse/native none rbind,create=file 0 0
lxc.mount.entry = /home/aleasto/.local/share/waydroid/data data none rbind 0 0
```

linsui: This file is generated correctly. The problem is that `/run/xdg` doesn't work for me.

PR #1330, opened 2024-03-19
Fix Networking issue

souta-lab:
https://github.com/waydroid/waydroid/issues/143 fixed this issue


Quackdoc:
this would just break for other users. This could be set behind a config flag.

PR #1345, opened 2024-04-01
initializer: Fix Halium version VNDK detection logic for 12L+

JamiKettunen:
API 32 (12L) vendor was getting detected as `HALIUM_13` which wouldn't work on devices when official images ship with support for these newer versions.


JamiKettunen:
@aleasto Perhaps you can check this, currently on an Android 12L vendor libhybris device it's detected as Halium 13 and so everything else is also offset by a version forward which definitely won't work

PR #1356, opened 2024-04-07
Add shutdown action

utku-helvaci:
`pkexec waydroid shell svc power shutdown` is not in the documentation and I've added it in worry of container corruption

### DIFF
--- a/data/Waydroid.desktop
+++ b/data/Waydroid.desktop
@@ -1,7 +1,13 @@
 [Desktop Entry]
-Type=Application
-Name=Waydroid
+Actions=Shutdown
+Categories=X-WayDroid-App;
 Exec=waydroid first-launch
 Icon=waydroid
-Categories=X-WayDroid-App;
+Name=Waydroid
+Type=Application
 X-Purism-FormFactor=Workstation;Mobile;
+
+[Desktop Action Shutdown]
+Exec=bash -c '(waydroid status|grep RUNNING) && if ! (pkexec waydroid shell svc power shutdown && waydroid session stop) && (waydroid status|grep RUNNING);then notify-send -i waydroid -a Waydroid '\\''Session has failed to shutdown !'\\'';fi'
+Icon=waydroid
+Name=Waydroid shutdown


JamiKettunen:
Have you experienced corruption with `waydroid session stop` or?

Lokawn:
The approach towards editing the desktop file will not work, as [this](https://github.com/waydroid/waydroid/blob/main/data/Waydroid.desktop) file is overidden on every launch of waydroid by entries created in .local/share/applications by [user_manager.py](https://github.com/waydroid/waydroid/blob/main/tools/services/user_manager.py).
This patch should work.
```
--- a/tools/services/user_manager.py
+++ b/tools/services/user_manager.py
@@ -61,6 +61,12 @@
 X-Purism-FormFactor=Workstation;Mobile;
 Icon=waydroid
 NoDisplay={str(hide).lower()}
+Actions=Shutdown
+
+[Desktop Action Shutdown]
+Exec=bash -c '(waydroid status|grep RUNNING) && if ! (pkexec waydroid shell svc power shutdown && waydroid session stop) ; then notify-send -i waydroid -a Waydroid \"Waydroid session has failed to shutdown\";fi'
+Icon=waydroid
+Name=Shutdown Waydroid
 """)
 
     def userUnlocked(uid):
```

TriMoon:
@JamiKettunen 
> Have you experienced corruption with `waydroid session stop` or?

Shutdown of the container is not same as Stopping the session :wink: 
But as @Lokawn pointed out this needs to be addressed in the `.py` file...

JMarcosHP:
Hello, what happened with the merger? Don't you want to include the patch for user_manager.py?

unique-EJ:
@JMarcosHP , @utku-helvaci , @TriMoon 

> This branch has conflicts that must be resolved

What happened?

... and if the Waydroid session shuts down, should the background process be stopped? (through `systemd`? `systemctl stop waydroid-container`). This will show a window to get authentication privileges.

TriMoon:
@unique-EJ no idea why you mentioned me, because im just another user just like you...

unique-EJ:
> @unique-EJ no idea why you mentioned me, because i'm just another [normal] user just like you...

@TriMoon sorry, I did not know that so I will not mention you here again.

TriMoon:
> > @unique-EJ no idea why you mentioned me, because i'm just another [normal] user just like you...
> 
> @TriMoon sorry, I did not know that so I will not mention you here again.

thanks, lets not keep on with this ping-pong then... :wink:

PR #1400, opened 2024-05-09
Re-enable nouveau and nvk

aleasto:


aleasto:
This won't hurt even if NVK isn't ready yet

PR #1422, opened 2024-05-26
require WM on `bug_report` template

Rudxain:
Many distros and DEs support multiple Window Managers/Systems. So it seems like a good idea to disambiguate which is running

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -52,6 +52,13 @@ body:
 
   - type: input
     attributes:
+      label: Window Manager / Windowing System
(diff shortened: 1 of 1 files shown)


Quackdoc:
this doesn't really make a bunch of sense since waydroid is wayland only, and compositor is the only thing that really matters since compositor and window management are almost always intertwined. Is there a case where this isnt true with wayland? though I suppose desktop environment should be changed to compositor regarding this

Rudxain:
> waydroid is wayland only

I thought it was "more cross-platform"!

> desktop environment should be changed to compositor

I agree, but I must admit I'm somewhat ignorant about this topic, so I'll wait for more reviews before pushing more commits

PR #1477, opened 2024-07-09
Add msm_dpu driver for freedreno

linsyking:
I am using a chromebook (trogdor model) which uses a Qualcomm turnip gpu. The driver name should be "msm_dpu" on my machine.


aleasto:
Sure, we can add this. Thanks.

PR #1579, opened 2024-09-29
systemd: add service file for 'waydroid session'

PureTryOut:
postmarketOS right now uses a desktop file in /etc/xdg/autostart to automatically start the Waydroid session upon login. Works well enough but this lacks proper service management, e.g. what to do when the service fails.
It's better to use systemd for this, it can handle auto-starting it and has proper logging infrastructure available.


Fuseteam:
i would love this this land

i previously tried to do this but i couldn't get it to launch properly after waydroid-container if and only if the container is ready: https://github.com/waydroid/waydroid/pull/678

Fuseteam:
hmmm seems to start too early on an ubuntu touch phone

Fuseteam:
huh seems my draft PR is incomplete too, i didn't include the parts needed to use dbus to activate the systemd service o.O

Fuseteam:
wait why this closed?

PureTryOut:
Because I forgot this repo exists, seeing it's been open without a meaningful response since forever not that strange, and cleared out some of the forked repositories I had.

I have no hope on landing this, ever, since upstream response is just non-existent.

Fuseteam:
dang, i do suppose the response is overdue. imma just ping @aleasto for his thoughts on this just in case, i deffo want this to land

PR #1606, opened 2024-10-30
Add svg icon of waydroid

### LINKED ISSUE #431: New svg icon for waydroid
Hi, I was working on the new svg icon of waydroid for an icon proposal which I found impossible in the png icon of waydroid and here is the output :-

AppIcon

Sorry, the icon looks damaged (outlined in white) due to the conversion of svg to png, can you try fixing it?
I am eager to submit my proposal in the icon design.

tehnick:
Closes: #431


ghost:
It includes original bitmap data inside :/ In terms of Wikimedia Commons - fake svg!

Twig6943:
> It includes original bitmap data inside :/ In terms of Wikimedia Commons - fake svg!

Seems to be a valid svg?

image

ghost:
My bad. It contains both svg and original bitmap data too. I would also say that vector data is dirty but it a separate issue... More important is bitmap data. Notice it is hidden inside pattern so it is not visible in normal "layers and objects" view

Twig6943:
Does this seem good?

AppIcon

image

ghost:
> Does this seem good?

It is probably better btw. But still very imperfect...


1.It doesn't contain unnecessary bitmap but it has unnecessary groups...
2. It has unnecessary `opacity` properties.
3. You probably don't need to have `id` properties
4. If you use groups you could uplift `fill` properties so they are inherited.
5. Better using `px` or just plain numbers instead of `pt` or for "space conservation" just delete `width` and `heigth` properties and leave only `viewBox`.
6. `DOCTYPE` isn't necessary as well as `xml` tag I usually leave xml but doctype doesn't have much sense nowadays except "strict compliance" so I advise to delete it. Inkscape doesn't use it for example and everyone is still happy
7. Looks like an export from ADOBE something or some affinity (don't remember which one adds the doctype and groups for everything)
8. Orange color looks more like original image




The point below is primarily venting of the person who knows svg and aims for the best perfection:

9. When I said it is not clear I primarily meant usage of lots of unnecessary points, usage of paths instead of plain shapes. It is questionable by many people and have some nasty side effects if you want to just have "wireframe" and print/cut it won't be so easy.... But I see that it would be possible to just use a circles and rects for creating main green objects. The capsules of antennas could even be do done by just one separate line with very think stroke... It quite depends on what you see as a clean svg... Also I hate the editors (apps) for loosing info about points. SVG has tons of [ways to do curves](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/Paths). Some of them can be hand made using special paths commands. For example for drawing the main green body you would need only 8 points. 4 points which constitute rect of each antenna... Or you could also need two points of middle of the image to cut the below part of circle. I am not sure which one would give you better size... But as the flat version is quite symmetric you could draw only half of the image and then mirror that half with just `<use id="<id_of_half>" transform="(<mirror + move>)">`...

PR #1681, opened 2025-01-23
support for user config from XDG_CONFIG_HOME

bhnasy15:
Added support for user config from XDG_CONFIG_HOME
Added property 'create_desktop_entry': only create desktop entries of True.
The function tools.config.load(args) has been changed to accept a file path.

### DIFF
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -25,6 +25,14 @@ config_keys = ["arch",
                "mount_overlays",
                "auto_adb"]
 
+property_keys = [
+        "create_desktop_entry",
+        ]
+
+property_defaults = {
+        "create_desktop_entry": "True",
+        }
+
 # Config file/commandline default values
 # $WORK gets replaced with the actual value for args.work (which may be
 # overridden on the commandline)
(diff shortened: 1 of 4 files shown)


FaintShadow:
> Added support for user config from XDG_CONFIG_HOME Added property 'create_desktop_entry': only create desktop entries of True. The function tools.config.load(args) has been changed to accept a file path.

Well I was having a discussion about that here #1687 and I think a more "complex" solution would be better. Do you wanna join us ?

bhnasy15:
sounds good.
for now, I added a "White List" option, the list along with the toggle should be good enough.
the config file would live in "$HOME/.config/waydroid/waydroid.cfg", and would look like this
```
[waydroid]
...

[properties]
create_desktop_entry = False
desktop_entry_white_list = "Files", "Contacts"
...
```

FaintShadow:
> sounds good. for now, I added a "White List" option, the list along with the toggle should be good enough. the config file would live in "$HOME/.config/waydroid/waydroid.cfg", and would look like this
> 
> ```
> [waydroid]
> ...
> 
> [properties]
> create_desktop_entry = False
> desktop_entry_white_list = "Files", "Contacts"
> ...
> ```

I think it would be better if we can chat about that in the discussion.

aleasto:
Waydroid config is something only the system administrator should have access to. Hence it is limited to `/var/lib/waydroid/waydroid.cfg`.

As for hiding desktop files, https://github.com/waydroid/waydroid/commit/018fb645c45e11833740dfdb943a9dd437c5e670 will now handle that better. Mind the the files must be created even if you don't want to see them, otherwise if you willingly or unwillingly launch such an application from the Android side, it wouldn't get an icon in your desktop environmnet. Just set `NoDisplay=true` to those you don't want to see.

FaintShadow:
> Waydroid config is something only the system administrator should have access to. Hence it is limited to `/var/lib/waydroid/waydroid.cfg`.
> 
> As for hiding desktop files, [018fb64](https://github.com/waydroid/waydroid/commit/018fb645c45e11833740dfdb943a9dd437c5e670) will now handle that better. Mind the the files must be created even if you don't want to see them, otherwise if you willingly or unwillingly launch such an application from the Android side, it wouldn't get an icon in your desktop environmnet. Just set `NoDisplay=true` to those you don't want to see.

So I have to go to EVERY SINGLE desktop entry that was created and simply do a `NoDisplay=true` ?

PR #1831, opened 2025-03-27
Check ipv6 enabled before trying anything

### LINKED ISSUE #586: eth0 initialized incorrectly when ipv6 is disabled
**Bug description**
When using the
`ipv6.disable=1`
kernel parameter, the initialization of the eth0 interface in the android system is performed incorrectly, including for ipv4. In the script `waydroid-net.sh` ipv6 settings are also not used. Addresses are not issued, respectively, the network does not work.
Also logs reported an error in 151 line of the `waydroid-net.sh`.

**General information:**
 - waydroid 1.3.3-1
 - image VANILLA and FOSS 17.1_20220723-1

**Desktop:**
 - OS: Arch
 - Kernel version: 6.0.7-zen1-1-zen
 
**Solution:**
After a few fun hours my solution is to change my own configuration to `ipv6.disable=0` and ignore version 6 packets using firewall.

Why could this happen and what is responsible for this bug? If this is an error in the network script, please fix it.

jflaflamme:
Closes #586

### DIFF
--- a/data/scripts/waydroid-net.sh
+++ b/data/scripts/waydroid-net.sh
@@ -151,7 +151,9 @@ start() {
     # set up the lxc network
     [ ! -d /sys/class/net/${LXC_BRIDGE} ] && ip link add dev ${LXC_BRIDGE} type bridge
     echo 1 > /proc/sys/net/ipv4/ip_forward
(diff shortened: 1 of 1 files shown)


aleasto:
I don't understand how this would change anything

aleasto:
I `/proc/sys/net/ipv6` does not exist, the `echo <> || true` command will simply return true. This patch does not change the behaviour.

PR #1897, opened 2025-05-23
data: Add a valid registered category to Waydroid.desktop for appstream

joebonrichie:
```
$ appstreamcli compose waydroid-install-prefix/
Automatically selected 'install/' as data output location.
WARNING: Metadata origin not set, using 'example'
Processing directory: install/
Composing metadata...
Run failed, some data was ignored.
Errors were raised during this compose run:
id.waydro.waydroid
  E: no-valid-category
Refer to the generated issue report data for details on the individual problems.
```

https://specifications.freedesktop.org/menu-spec/latest/category-registry.html

This fixes composing waydroid into a catalog when using appstream. Without it waydroid will fail to be included in the catalog for distros that use `appstream-generator`.


aleasto:
Thank you.

I've applied the same change to the runtime-generated desktop file.

PR #1898, opened 2025-05-24
Remove unused python imports

ddogfoodd:
> [!WARNING]  
> I didn't actually test the changes, so a test/review would be very much appreciated and required before a potential merge.

remove some unused python imports to reduce memory usage slightly


aleasto:
Thanks, lgtm

PR #1971, opened 2025-07-14
first-launch: Update to GTK 4

linkmauve:
There has been no functional change to this UI, besides updating to the newest version of the toolkit.

There is a CRITICAL error displayed because the application id should be a reverse.dns.Name, but this would also require renaming the icon to the same name, so for now it isn’t done.

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -284,8 +284,8 @@ def remote_init_server(args, params):
 def remote_init_client(args):
     # Local imports cause Gtk is intrusive
     import gi
-    gi.require_version("Gtk", "3.0")
-    from gi.repository import Gtk
+    gi.require_version("Gtk", "4.0")
+    from gi.repository import Gtk, Gio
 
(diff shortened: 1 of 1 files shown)


aleasto:
> There is a CRITICAL error displayed because the application id should be a reverse.dns.Name

Yeah because using GApplication is overkill here. Is that mandatory with GTK4? If so, can we at least use `G_APPLICATION_NON_UNIQUE` ?

linkmauve:
`Gtk.Application` is indeed mandatory now that `Gtk.main()` has been removed, all GTK 4 applications use it.

I have replaced the wrong `application_id` with `Gio.ApplicationFlags.NON_UNIQUE`, to disable the DBus name registration done by `Gio.Application`, it removed the warning and still displays the correct icon.

aleasto:
Thanks, looks better.

However Waydroid runs on some pretty old distributions which may not ship gtk4, including debian bullseye.
I'm afraid this should wait until distributions without gtk4 are EOL.

PR #2089, opened 2025-10-14
Remove underscores from .desktop action names

korbel:
According to the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/1.2/extra-actions.html#extra-actions-identifier), identifiers for additional actions in desktop files must follow the same format as key names — only the characters `A–Z`, `a–z`, `0–9`, and `-` are allowed.

Currently, the generated desktop files use underscores in the action identifiers, which causes `desktop-file-utils` validation to fail. Replacing underscores with hyphens resolves the issue.


aleasto:
Thanks!

Issue #2116, opened 2025-11-08
[BUG] `1.5.4-5.fc43` fails to install GSIs from `android.settings.development.DSULoader` because `DynamicSystemService` fails to create `/userdata`.


RokeJulianLockhart:
#### How To Reproduce

~~~sh
#!/usr/bin/sh
adb shell cmd activity start --user 0 -a android.intent.action.VIEW -n com.android.settings/com.android.settings.development.DSULoader -f 0 com.android.settings && \
adb exec-out uiautomator dump /dev/tty \
| sed 's/></>\n</g' \
| awk '
/text="GSI\+GMS X86_64"/ {
    match($0, /bounds="\[([0-9]+),([0-9]+)\]\[([0-9]+),([0-9]+)\]/, t);
    if (t[1]) {
        minx=t[1]; miny=t[2]; maxx=t[3]; maxy=t[4];
        # Now look upward to find clickable parent
        for (i=NR-1;i>NR-20;i--) {
            if (buf[i] ~ /clickable="true"/ && match(buf[i], /bounds="\[([0-9]+),([0-9]+)\]\[([0-9]+),([0-9]+)\]/, c)) {
                cx=(c[1]+c[3])/2; cy=(c[2]+c[4])/2;
                system("adb shell input tap " cx " " cy);
                exit
            }
        }
    }
}
{ buf[NR]=$0 }' && \
adb exec-out uiautomator dump /dev/tty \
| sed 's/></>\n</g' \
| awk '
/text="Agree"/ && /clickable="true"/ {
    if (match($0, /bounds="\[([0-9]+),([0-9]+)\]\[([0-9]+),([0-9]+)\]/, b)) {
        cx=(b[1]+b[3])/2;
        cy=(b[2]+b[4])/2;
        system("adb shell input tap " cx " " cy);
        exit
    }
}'
~~~

#### The Consequence

With irrelevant noise removed, `logcat` returns what [`gist.github.com/RokeJulianLockhart/3f36d04c6ca28f4b580300caf91b64a2`](https://gist.github.com/RokeJulianLockhart/3f36d04c6ca28f4b580300caf91b64a2/revisions?short_path=ec89c94#diff-ec89c9458475650db8037784dceb43e21bac041dae51a0a88b9d2862472c51ad) contains. Immediately thereafter, `adb shell dumpsys notification --noredact` [^1]: returns:

[...]

PR #2117, opened 2025-11-08
Update the `bug_report` template to be easier to utilise.

RokeJulianLockhart:
1.	To make collecting logs as easy as is possible, the commands now invoke the relevant files in the default text area.

	However, if they fail to invoke, the text files still remain, because this section has merely been appended to existent commands.

	The mere file paths that have been converted into commands remain readable.

2.	`input`s have become `textarea`s, so that users can enter sufficient information without butchering it. [`issues/2116`](https://github.com/waydroid/waydroid/issues/2116#issue-3603657294) is an example. This shouldn't impact most users.

3.	Removed redundant quotation from the “Logs” string, solely to ensure that quotation is consistent across the document.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -10,7 +10,7 @@ body:
     validations:
       required: true
 
-  - type: input
+  - type: textarea
     attributes:
       label: Waydroid version
       description: Output of `waydroid --version`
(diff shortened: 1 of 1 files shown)


aleasto:
> `input`s have become `textarea`s, so that users can enter sufficient information without butchering it.

It is intended that these be one-line text entries. One line is sufficient for all the questions asked.

beedell-roke:
@aleasto, in `etc/os-release`, the name of the OS, it's version, and variant are all on separate lines, and `ini` files can't be made into single lines. For KDE's `kinfo`, and Windows's `Get-Computer` (where WSL is relevant), the output is still YAML, which *can*, of course, be made single-line. However, it's multi-line by default.

aleasto:
Going to reject this. I implemented a `waydroid bugreport` command instead that gathers all information

beedell-roke:
@aleasto, why are they mutually exclusive? People shall still manually fill the form. Is it because you wouldn't want to refactor the command to match any changed template, perhaps?

Issue #2120, opened 2025-11-10
[BUG] Python script failed during installation on Ubuntu Jammy


sunflower2333:
### Describe the bug

This [commit](https://github.com/waydroid/waydroid/commit/a635cf076697dc3faa6776119146278c5c3752f8) cause a f string error while installing waydroid.

The Python interpreter may interpret the quotes as the end of the F string.

### Waydroid version

latest

### Device

Linux Desktop

### Operating System

Ubuntu 22.04 arm64

### Kernel version

6.18-rc4

### Desktop Environment

kde-plasma

### GPU

Adreno 740

### Logs

Logs:
```
Setting up waydroid (1.5.4) ...
Traceback (most recent call last):
  File "/usr/bin/waydroid", line 7, in <module>
    import tools
  File "/usr/lib/waydroid/tools/__init__.py", line 168
    f"  {sys.argv[0]} --details-to-stdout {" ".join(sys.argv[1:])}")
                                                                   ^
SyntaxError: f-string: expecting '}'
  File "/usr/lib/waydroid/tools/__init__.py", line 168
    f"  {sys.argv[0]} --details-to-stdout {" ".join(sys.argv[1:])}")
                                                                   ^
SyntaxError: f-string: expecting '}'
dpkg: error processing package waydroid (--configure):
 installed waydroid package post-installation script subprocess returned error exit status 1
```

mags0ft:
I was able to reproduce the same bug. Fails for me too, `sudo apt update` is fine but `sudo apt upgrade` fails.

The more [I look at it](https://github.com/waydroid/waydroid/blob/7b19c643b29f6bd8516593f6fc5efbde5e434dcf/tools/__init__.py#L168), the less I find the error though. Isn't that valid syntax? What's happening here??

Edit: btw, haven't tried but this is a wild guess, this should be fixable by running `sudo nano /usr/lib/waydroid/tools/__init__.py` and commenting out lines 166 to 168, then running `sudo apt upgrade` again. Use with caution, I haven't tested it yet.

Edit: Oh, I see, Python may be confusing the `"` before the `.join(` part as the end of the f-string. Wasn't this patched for newer Python versions? I think this may be related to it. I am running 3.10.12.

sunflower2333:
[...]

PR #2121, opened 2025-11-10
Fix: replace '"' that may cause SyntaxErrors for some Python versions

mags0ft:
Hello there,

As mentioned in [this issue](https://github.com/waydroid/waydroid/issues/2120), some installs or updates are suddenly failing.

This patch replaces the ambiguous `"` at the `.join(` part of line 168 causing install failures for a.e. some Ubuntu versions with specific Python versions which fail to interpret the syntax correctly.

Tested on Ubuntu 22.04 LTS, Python 3.10.

Have a great day!


aleasto:
Thanks!

PR #2131, opened 2025-11-22
http: Standardize file size and download speed units

Integral-Tech:
Update the representations of file sizes and download speeds to correct the informal units (kb, mb, kbps and mbps):

- Changed kb to kB for kilobytes
- Changed mb to MB for megabytes
- Changed kbps to kB/s for kilobytes per second
- Changed mbps to MB/s for megabytes per second


aleasto:
Sounds good. Thanks!

PR #2136, opened 2025-11-25
Also consider block device files when checking for preinstalled images

RushingAlien:
Hey, so I was trying to `sudo waydroid init  -f` where  both preinstalled system.img and vendor.img are symlinks and the initializer tool just refuses to acknowledge it exists. My use case is irunning Waydroid in a VM and to reduce overhead i had the system.img and vendor.img files in the host's file system where both files are passed to Qemu as disk images. These disks are then symlinked as system.img and vendor.img in the image path, as seen below:
```
ls /etc/waydroid-extra/images/ -l
total 8
lrwxrwxrwx 1 root root 8 Nov 25 21:01 system.img -> /dev/sda
lrwxrwxrwx 1 root root 8 Nov 25 21:01 vendor.img -> /dev/sdb
```
Had to 
1. create empty stub/spoof files
2. `waydroid init -f`
3. replace stub files with links
4. start waydroid
5. works

This PR should remove the need for such a workaround.

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -52,9 +52,9 @@ def setup_config(args):
         if os.path.isdir(preinstalled_images):
             system_path = preinstalled_images + "/system.img"
             vendor_path = preinstalled_images + "/vendor.img"
-            are_files = os.path.isfile(system_path) and os.path.isfile(vendor_path)
-            are_blocks = stat.S_ISBLK(os.stat(system_path).st_mode) and stat.S_ISBLK(os.stat(vendor_path).st_mode)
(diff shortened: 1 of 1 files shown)


aleasto:
That's not quite right. We don't want to accept symlinks to a directory for example.

`isfile()` follows symlinks, so the check fails in your case not because the path is a symlink, but because it is a symlink to a *block device* rather than to a regular file; use `pathlib.Path.is_block_device()` to check for that.

RushingAlien:
Thanks for pointing that out.

aleasto:
Thanks!

PR #2144, opened 2025-12-04
gpu: add support for xe kernel driver

arthefaz:
Adds support for newer Intel graphics device using the `xe` driver


aleasto:
Thanks

PR #2212, opened 2026-01-29
Support OpenRC init system

GermanAizek:
There are many users with this initialization system
Main repo: https://github.com/OpenRC/openrc
More info: https://en.wikipedia.org/wiki/OpenRC


Quackdoc:
for supporting multiple service systems, which I do like, perhaps we should make a services/<service-manager> folder?

PR #2241, opened 2026-03-01
tools: Fix all occurrences of unused variable warning

arkq:
This PR also fixes the default branch name in the "Check Python Package" GitHub Action workflow file, so the static analysis will be run on every PR to the `main` branch.


aleasto:
Thanks

PR #2244, opened 2026-03-04
session: Add opt-in idle monitor to auto-stop sessions

### LINKED ISSUE #1052: Stop / restart the session from within the Waydroid UI
**Is your feature request related to a problem? Please describe.**
1. I'd like to be able to exit from Waydroid and stop the running session from within the Waydroid UI itself, without the need to run `waydroid session stop` externally. Currently, when you close the Waydroid window, the session keeps running in the background (see #774).

2. Restarting the session would also be nice, but I understand it's more complicated, so that's optional.

**Describe the solution you'd like**
1. I think it's a good idea to have this feature in Android power menu. Currently, there is no way to open it.

2. I can run Reboot from the Magisk UI, and Waydroid seems to shut down, but not completely – the session keeps running, and the restart doesn't happen. It would be great if Reboot could actually work the same way as `waydroid session stop && waydroid session start`.

**Describe alternatives you've considered**
1. Switching to a terminal window and issuing `waydroid session stop`.
2. Switching to a terminal window and issuing `waydroid session stop && waydroid session start`.

### LINKED ISSUE #2234: [BUG]Waydroid doesn't stop session when shutting down, get Waiting for binder Service Manager...
### Describe the bug

Works fine on first boot, choose shutdown in waydroid, shuts down and can not restart.

Waiting for binder Service Manager...
Service Manager never appeared
Failed to access IPlatform service

Reboot and it works fine again for the 1st run.

After doing some testing, I found the session never stops, so if I use the following command I can then start waydroid again for 1 time.

waydroid -v session stop

### Waydroid version

1.6.1

### Device

Linux Desktop

### Operating System

Ubuntu 25.10

### Kernel version

6.17.0-14-generic

### Desktop Environment

GNOME Shell 49.0

### GPU

AMD Radeon Vega 6

### Logs

/var/lib/waydroid/waydroid.cfg
/var/lib/waydroid/waydroid.prop
/var/lib/waydroid/waydroid_base.prop
/var/lib/waydroid/waydroid.log
`sudo waydroid shell -- logcat -d | tee logcat.txt`
`sudo dmesg | tee dmesg.txt`

countgitmick:
When `stop_on_idle=True` is set in `waydroid.cfg`, the session periodically checks whether the container is still running and whether there are active user tasks. After 30 seconds of idle state, the session stops itself.

This fixes two long-standing issues:
- The session not stopping when Android shuts down from its own UI (#2234) — detected via D-Bus container state check, works immediately
- The session not stopping when all app windows are closed (#1052) — detected via a new `getRunningTasks` binder transaction, requires the companion Android-side change

The idle monitor uses `GLib.timeout_add_seconds()` inside the existing mainloop. No new threads or dependencies. A non-blocking `get_service_nowait()` is added to `IPlatform` to avoid stalling the main loop.

Enable: add `stop_on_idle=True` to the `[waydroid]` section of `/var/lib/waydroid/waydroid.cfg`

Note: #1052 requires both this PR and the companion Android-side PR to be merged (new Android image built). #2234 is fully fixed by this PR alone.

Companion PR: waydroid/android_vendor_waydroid#48

Fixes #1052
Fixes #2234

### DIFF
--- a/tools/actions/session_manager.py
+++ b/tools/actions/session_manager.py
@@ -7,6 +7,7 @@ import sys
 import tools.config
 import tools.helpers.ipc
 from tools import services
+from tools.interfaces import IPlatform
 import dbus
 import dbus.service
 import dbus.exceptions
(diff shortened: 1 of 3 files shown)


aleasto:
This is already achieved by `suspend_action=stop`, without polling

countgitmick:
You're right that suspend_action=stop handles the case where Android initiates shutdown through its own power management. The gap this addresses is when all app windows are closed but the container hasn't triggered a suspend. The session hangs indefinitely (issues #2234, #1052). In that state, suspend_action=stop won't fire because Android doesn't consider itself sleeping.

An event-driven approach (a Binder callback when the last task ends) would be ideal, but that would require changes on the Android image side. The lightweight poll via getRunningTasks() was chosen as a pragmatic solution that works with existing images. It depends on what you think is the better approach here. Also I could be missing an event that I'm unaware of. Would be glad to get your feedback on this.

PR #2248, opened 2026-03-11
tools: Fix all occurrences of catching bare exceptions

arkq:
Also added `check=True` when running `sudo` to check for incorrect password.


aleasto:
Thanks!

PR #2250, opened 2026-03-13
Add show_boot_animation config key to control boot animation during app launch and intent launch

sidevesh:


MeechaPooch:
Awesome!

Issue #537, opened 2022-09-29
Configure to work without listen port 53


itsnotsaved:
**Describe the Problem**

I install waydroid on ubuntu machine using https://github.com/n1lby73/waydroid-installer and i saw it listen to port 53 with dnsmasq; how to disable that? I use a local DNS server (adguradhome; what is it? similar to pihole etc) on my laptop and it needs the port 53; also i don't like waydroid use its own way to resolve DNS (no such trouble with anbox as i remember)

I couldn't find any details on https://docs.waydro.id

**General information (please complete the following information):**
 - Waydroid tools Version - 1.3.2
 - Waydroid Images Version [above]

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.01 64 bit
 - GPU: Skylake GT2
 - Kernel version: 5.15.0-48-generic
 - Desktop: Gnome 42.4

**Smartphone (please complete the following information):**

I don't use

Quackdoc:
by changing the waydroid NIC in waydroid's LXC config, you can create your own network interface and rules, it's not something I yet have documented.

itsnotsaved:
@Quackdoc please add a guide for it

Important parts:
- Free the port 53
- Configure to resolve DNS through custom DNS (e.g. LAN DNS IP) or system DNS or external DNS

acheong08:
Updates? I have something running on port 53 that is conflicting with dnsmasq

Petemir:
Ok, got it working for my case. 

The `/usr/lib/waydroid/data/scripts/waydroid-net.sh` script starts `dnsmasq` on the host. `dnsmasq` provides both DNS (port 53) and DHCP (port 67) servers to the waydroid container; this will conflict with any local DNS/DHCP servers running in the host. As I am running `pihole` through Docker, my default docker-composer.yml file was also forwarding port 67, even though I didn't use it. So, my first step was totally dependent on me using `pihole`:

* Disable port 67 forwarding in my docker-compose.yml file and recreate the pihole container.

[...]

PR #2279, opened 2026-04-18
waydroid-net: allow overriding dnsmasq conf path

### LINKED ISSUE #2278: waydroid-net: expose dnsmasq conf-file override without patching the installed script
## Summary

`data/scripts/waydroid-net.sh` already supports passing a custom dnsmasq config file through `LXC_DHCP_CONFILE`, but there is currently no supported way to set it without editing the packaged script.

This makes a number of host networking setups depend on local patches to `/usr/lib/waydroid/data/scripts/waydroid-net.sh`.

## Current behavior

Today the script hardcodes:

```sh
LXC_DHCP_CONFILE=""
```

and later consumes it here:

```sh
LXC_DHCP_CONFILE_ARG="--conf-file=${LXC_DHCP_CONFILE:-/dev/null}"
```

So the implementation already has the "custom conf file" capability internally, but users/distributions do not have a stable override point for it.

## Why this seems worth fixing

- Users are already editing `waydroid-net.sh` by hand in the field to change dnsmasq behavior.
- Related issue: #537
- There was also a previous networking flexibility PR: #670
- In practice this affects setups that need custom upstream DNS, custom DHCP DNS options, temporary logging, or integration with host-side DNS/proxy stacks.

This is not meant to change the default networking behavior for normal installs. It is only about exposing an already-existing internal knob in a supported way.

## Proposed direction

Allow an optional environment override, for example:

```sh
: "${LXC_DHCP_CONFILE:=${WAYDROID_DNSMASQ_CONF:-}}"
```

That keeps the current default intact:

- unset override -> still uses `--conf-file=/dev/null`
- override set -> callers can opt in to an external dnsmasq config file

This would let admins or distributions configure it with a systemd drop-in such as:

```ini
[Service]
Environment="WAYDROID_DNSMASQ_CONF=/etc/waydroid-dnsmasq.conf"
```

without having to patch the installed `waydroid-net.sh` file after every upgrade.

## Why environment first

I think this is the smallest possible change:

- no default behavior change
- no CLI change
- no config migration
- no new mandatory dependency between Python config parsing and the shell script

If maintainers would prefer a `waydroid.cfg` key instead, that also sounds reasonable, but the environment override seems like the narrowest first step.

LyHug:
## Summary

Allow `waydroid-net.sh` to optionally inherit a dnsmasq config path from `WAYDROID_DNSMASQ_CONF`, while keeping the current default behavior unchanged.

Closes #2278

## Why

The script already supports `LXC_DHCP_CONFILE` internally and later maps it to:

```sh
--conf-file=${LXC_DHCP_CONFILE:-/dev/null}
```

but there is currently no supported external override for it. As a result, users who need custom dnsmasq behavior end up patching the installed `waydroid-net.sh` script directly.

Related discussion:

- #537
- #670

## Behavior

When `WAYDROID_DNSMASQ_CONF` is unset, behavior stays the same as today:

- `LXC_DHCP_CONFILE` resolves to empty
- dnsmasq still gets `--conf-file=/dev/null`

When it is set, callers can opt in to an external dnsmasq config file without modifying the packaged script.

## Example

```ini
[Service]
Environment="WAYDROID_DNSMASQ_CONF=/etc/waydroid-dnsmasq.conf"
```

## Validation

- `sh -n data/scripts/waydroid-net.sh`

### DIFF
--- a/data/scripts/waydroid-net.sh
+++ b/data/scripts/waydroid-net.sh
@@ -22,7 +22,9 @@ LXC_NETMASK="255.255.255.0"
 LXC_NETWORK="192.168.240.0/24"
 LXC_DHCP_RANGE="192.168.240.2,192.168.240.254"
 LXC_DHCP_MAX="253"
-LXC_DHCP_CONFILE=""
+# Allow an external dnsmasq config to be injected without changing the
+# default hermetic /dev/null behavior.
+: "${LXC_DHCP_CONFILE:=${WAYDROID_DNSMASQ_CONF:-}}"
 LXC_DHCP_PING="true"
 LXC_DOMAIN=""
 LXC_USE_NFT="false"
--- a/data/scripts/waydroid-net.sh
+++ b/data/scripts/waydroid-net.sh
@@ -199,7 +199,7 @@ start() {
         mkdir "${varlib}"/misc
     fi
 
-    dnsmasq $LXC_DHCP_CONFILE_ARG $LXC_DOMAIN_ARG $LXC_DHCP_PING_ARG -u ${DNSMASQ_USER} \
+    dnsmasq "$LXC_DHCP_CONFILE_ARG" $LXC_DOMAIN_ARG $LXC_DHCP_PING_ARG -u ${DNSMASQ_USER} \
             --strict-order --bind-interfaces --pid-file="${varrun}"/dnsmasq.pid \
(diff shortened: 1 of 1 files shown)


LyHug:
Good catch on the shell word-splitting case. I've pushed 3b12a97, which quotes the dnsmasq conf-file argument so paths containing spaces are passed as a single argument.

PR #2285, opened 2026-04-23
Toggle desktop file

max-4-3:
# Feature
Added a feature to toggle creation of `waydroid.*.desktop` files  
It can be a bit annoying that waydroid recreate those files every time it boots and they come up in search like `krunner`  

## HOW
Add ( or change ) `no_applications` config key in `/usr/lib/waydroid/waydroid.cfg` to:
- True: Disable creation of those files ( and additionally remove them if present )
-  False: Enable creation of those files ( and additionally create them if not present )

### example:
`/var/lib/waydroid/waydroid.cfg`:
```cfg
# other configs...
no_applications = True
# other configs...
```

### DIFF
--- a/tools/config/__init__.py
+++ b/tools/config/__init__.py
@@ -23,7 +23,8 @@ config_keys = ["arch",
                "vendor_datetime",
                "suspend_action",
                "mount_overlays",
-               "auto_adb"]
+               "auto_adb",
+               "no_applications"]
 
 # Config file/commandline default values
 # $WORK gets replaced with the actual value for args.work (which may be
@@ -43,6 +44,7 @@ defaults = {
     "auto_adb": "False",
     "container_xdg_runtime_dir": "/run/xdg",
     "container_wayland_display": "wayland-0",
+    "no_applications": "False"
 }
 defaults["images_path"] = defaults["work"] + "/images"
 defaults["rootfs"] = defaults["work"] + "/rootfs"
--- a/tools/helpers/arguments.py
+++ b/tools/helpers/arguments.py
@@ -15,7 +15,7 @@ import tools.config
     the "args" variable, which is prominently passed to most functions all
     over the waydroid code base.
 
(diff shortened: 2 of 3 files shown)


Twig6943:
You might need to make it so that it creates the .desktop files but adds in `NoDisplay=true`

From earlier PR:
https://github.com/waydroid/waydroid/pull/1083

aleasto:
> It can be a bit annoying that waydroid recreate those files every time it boots

Yes, they're not optional.
If you don't want to see an application, hide it with No display=true

I'm not interested in adding a configuration option for this.

Thank you.

max-4-3:
oh shoot i didn't knew about that ;-;

Twig6943:
Waydroid generates a lot of .desktop files over time and having to open each of them and adding in NoDisplay=true is kinda time consuming. Or running a sed command thats like `sed waydroid-* > NoDisplay=True` each time. I don't know why you're against the idea of adding in an option to add in NoDisplay=True to all .desktop files

aleasto:
Because the goal of Waydroid is to transparently handle android apps as desktop apps.
We're not quite there yet, but such an option goes in the opposite direction.

max-4-3:
i think that user should have an option to toggle that behavior about desktop file.  
because when i use my machine for long time i tend to quickly type app name and press enter ( in krunner ) and mostly ( and i mean mostly ) it open the wrong app ( waydroid one ) and now i have to make a run to my terminal to stop/kill the container and session ( so it doesn't slow down my entire system ;-; ).  
I assume my changes were too "distructive" lol  
But i like it, and i am keeping this feature locally and yeah.

Twig6943:
> i think that user should have an option to toggle that behavior about desktop file. because when i use my machine for long time i tend to quickly type app name and press enter ( in krunner ) 

I opened the wrong settings app at least 20 times by now xD

max-4-3:
for me it was telegram ;-;

PR #2290, opened 2026-04-26
protocol: Use aidl5/6 for service manager on API 35/36

supechicken:
Requires libgbinder 1.1.45 or above ([waydroid/libgbinder#10](https://github.com/waydroid/libgbinder/pull/4))

Tested with `waydroid app list` on A16


aleasto:
Thank you

PR #2310, opened 2026-05-09
gpu: Remove radeon from Vulkan driver map

supechicken:
The RADV driver only supports `amdgpu` kernel driver:

> Note that for GFX6-7 (GCN 1-2) GPUs, the amdgpu kernel driver is currently not the default in Linux (by default the old radeon KMD is used for these old GPUs, which is not supported by RADV)


aleasto:
Thanks

PR #2321, opened 2026-05-20
lxc: Generate dalvik props

EliasTheBigMasterofIssues:
Depends [!15](https://github.com/waydroid/android_device_waydroid_waydroid/pull/15)

Generate DEX2OAT properties based on cpus core count alongside with generating VM properties based on physical RAM

### DIFF
--- a/tools/helpers/lxc.py
+++ b/tools/helpers/lxc.py
@@ -352,14 +352,17 @@ def make_base_props(args):
         props.extend(dalvik_dex2oat)
     except Exception as e:
         logging.info(f"Error: Can't be possible to generate Dalvik's DEX2OAT properties: {e}")
-    
+
     # Generate necessary dalvik_vm props based on total physical ram
     try:
+        dalvik_vm = []
+        host_arch = True if "64bit" in platform.architecture() else False
+        if host_arch:
+            dalvik_vm.append("dalvik.vm.dex2oat64.enabled=true")
         mem_bytes = os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')
         mem_gib = round(mem_bytes / (1024.**3))
-        dalvik_vm = []
         if mem_gib >= 16:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=32m",
                 "dalvik.vm.heapgrowthlimit=448m",
                 "dalvik.vm.heapsize=640m",
@@ -368,7 +371,7 @@ def make_base_props(args):
                 "dalvik.vm.heaptargetutilization=0.4",
             ]
         elif mem_gib >= 12:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=24m",
                 "dalvik.vm.heapgrowthlimit=384m",
                 "dalvik.vm.heapsize=512m",
@@ -377,7 +380,7 @@ def make_base_props(args):
                 "dalvik.vm.heaptargetutilization=0.42",
             ]
         elif mem_gib >= 8:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=24m",
                 "dalvik.vm.heapgrowthlimit=256m",
                 "dalvik.vm.heapsize=512m",
@@ -386,7 +389,7 @@ def make_base_props(args):
                 "dalvik.vm.heaptargetutilization=0.46",
             ]
         elif mem_gib >= 6:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=16m",
                 "dalvik.vm.heapgrowthlimit=256m",
                 "dalvik.vm.heapsize=512m",
@@ -395,7 +398,7 @@ def make_base_props(args):
                 "dalvik.vm.heaptargetutilization=0.5",
             ]
         elif mem_gib >= 4:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=8m",
                 "dalvik.vm.heapgrowthlimit=256m",
                 "dalvik.vm.heapsize=512m",
@@ -404,7 +407,7 @@ def make_base_props(args):
                 "dalvik.vm.heaptargetutilization=0.6",
             ]
         elif mem_gib >= 2:
-            dalvik_vm = [
+            dalvik_vm += [
                 "dalvik.vm.heapstartsize=8m",
                 "dalvik.vm.heapgrowthlimit=192m",
                 "dalvik.vm.heapsize=512m",
@@ -416,7 +419,7 @@ def make_base_props(args):
             logging.info("Your physical RAM is lower than 2GB: Dalvik's VM properties can't be generated")
         props.extend(dalvik_vm)
     except Exception as e:
-        logging.info(f"Error: Can't be possible to generate Dalvik's VM properties: {e}")    
+        logging.info(f"Error: Can't be possible to generate Dalvik's VM properties: {e}")
 
     # Detect if System has Low Memory RAM
     try:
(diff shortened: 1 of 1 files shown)


tools/helpers/lxc.py
```diff
@@ -337,6 +337,84 @@ def find_hidl(intf):
...
+    for cpus in range(cpus_count):
+        list_cpu.append(str(cpus))
+        cpus_range = ','.join(list_cpu)
+    count = 0
+    for prop in dalvik_dex2oat:
+        if "cpu-set" in prop:
+            dalvik_dex2oat[count] += f"={cpus_range}"
+        elif "threads" in prop:
+            dalvik_dex2oat[count] += f"={cpus_count}"
+        count+=1
```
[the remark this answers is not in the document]
EliasTheBigMasterofIssues: i think the chance to cpu_count return none is zero, waydroid already run as root.
about cpus_range, it's is necessary, everything with "cpu-set=" depends to receive a list with all cpus
like if are 2 cores, it needs to be setup "dalvik.vm.background-dex2oat-cpu-set=0,1"

tools/helpers/lxc.py
```diff
@@ -337,6 +337,84 @@ def find_hidl(intf):
...
+       dalvik_vm[3]+="=8m"
+       dalvik_vm[4]+="=56m"
+       dalvik_vm[5]+="=0.42"
+    elif mem_gib >= 16:
+       dalvik_vm[0]+="=32m"
+       dalvik_vm[1]+="=448m"
+       dalvik_vm[2]+="=640m"
+       dalvik_vm[3]+="=16m"
+       dalvik_vm[4]+="=64m"
+       dalvik_vm[5]+="=0.4"
```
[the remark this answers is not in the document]
EliasTheBigMasterofIssues: if i copy this code will be necessary to add gemini as coauthor?

tools/helpers/lxc.py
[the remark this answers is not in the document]
EliasTheBigMasterofIssues: sorry, i'm understading now, is necessary to add gemini as coauthor?

EliasTheBigMasterofIssues:
Rebased

EliasTheBigMasterofIssues:
Sorry my editor has changed code format, i will fix it later on a rebase

EliasTheBigMasterofIssues:
Fixed

tools/helpers/lxc.py
```diff
@@ -337,6 +337,100 @@ def find_hidl(intf):
...
+        dalvik_dex2oat = [
+            f"dalvik.vm.background-dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.background-dex2oat-threads={cpus_count}",
+            f"dalvik.vm.boot-dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.boot-dex2oat-threads={cpus_count}",
+            f"dalvik.vm.dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.dex2oat-threads={cpus_count}",
+        ]
+        props.extend(dalvik_dex2oat)
+    except Exception as e:
```
aleasto: What is this meant to catch?

tools/helpers/lxc.py
```diff
@@ -337,6 +337,100 @@ def find_hidl(intf):
...
+    try:
+        cpus_count = os.cpu_count()
+        cpus_range = ",".join(map(str, range(cpus_count)))
+        dalvik_dex2oat = [
+            f"dalvik.vm.background-dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.background-dex2oat-threads={cpus_count}",
+            f"dalvik.vm.boot-dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.boot-dex2oat-threads={cpus_count}",
+            f"dalvik.vm.dex2oat-cpu-set={cpus_range}",
+            f"dalvik.vm.dex2oat-threads={cpus_count}",
```
aleasto: Is there any point in setting these to the full CPU count? What's the default behaviour? I would think that the intended use-case for these configs is to limit the cores where dex2oat can run, for to limit it to efficient cores in big.LITTLE architectures.

tools/helpers/lxc.py
```diff
@@ -337,6 +337,100 @@ def find_hidl(intf):
...
+                "dalvik.vm.heapgrowthlimit=192m",
+                "dalvik.vm.heapsize=512m",
+                "dalvik.vm.heapminfree=512k",
+                "dalvik.vm.heapmaxfree=8m",
+                "dalvik.vm.heaptargetutilization=0.75",
+            ]
+        else:
+            logging.info("Your physical RAM is lower than 2GB: Dalvik's VM properties can't be generated")
+        props.extend(dalvik_vm)
+    except Exception as e:
```
aleasto: What is this meant to catch?

tools/helpers/lxc.py
```diff
@@ -337,6 +337,100 @@ def find_hidl(intf):
...
+        logging.info(f"Error: Can't be possible to generate Dalvik's VM properties: {e}")
+
+    # Detect if System has Low Memory RAM
+    try:
+        mem_bytes = os.sysconf('SC_PAGE_SIZE') * os.sysconf('SC_PHYS_PAGES')
+        mem_gib = round(mem_bytes / (1024.**3))
+        if mem_gib <= 5:
+            logging.info("Low Memory Ram detected: Adding Low Memory RAM property")
+            props.append("ro.config.low_ram=true")
+    except Exception as e:
```
aleasto: What is this meant to catch?

tools/helpers/lxc.py
```diff
@@ -337,6 +337,100 @@ def find_hidl(intf):
...
+        ]
+        props.extend(dalvik_dex2oat)
+    except Exception as e:
+        logging.info(f"Error: Can't be possible to generate Dalvik's DEX2OAT properties: {e}")
+
+    # Generate necessary dalvik_vm props based on total physical ram
+    try:
+        dalvik_vm = []
+        host_arch = True if "64bit" in platform.architecture() else False
+        if host_arch:
```
aleasto:
```suggestion
        if platform.architecture()[0] == "64bit":
```

tools/helpers/lxc.py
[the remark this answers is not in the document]
aleasto: no :)

aleasto (changes requested):
Thanks. Overall, I agree with the idea.
Left some comments in-line.

tools/helpers/lxc.py
EliasTheBigMasterofIssues: Sorry for delay, i very newbie, i have think about try to use as maximum to compile when system are initializing, but can be better if being a half like most of phones uses? sorry

tools/helpers/lxc.py
EliasTheBigMasterofIssues: i just have used a suggestion from Gemini reviewer, but appears to be a false positive 
>These include a potential crash if the CPU count is unavailable, a critical bug resulting in invalid property >entries for systems with less than 2 GiB of RAM, and a missing heap size definition for the 2-4 GiB range. >Additionally, the feedback highlights PEP 8 indentation violations and suggests refactoring the logic to avoid >fragile manual index tracking and inefficient string construction.

tools/helpers/lxc.py
EliasTheBigMasterofIssues: is like same from gemini reviewer, but appears to be a false positive, i can remove it?

tools/helpers/lxc.py
EliasTheBigMasterofIssues: sorry is just false positive from gemini, i can remove all try catch?

tools/helpers/lxc.py
EliasTheBigMasterofIssues: **** i think this is fixed, i'am dividing per 2

tools/helpers/lxc.py
EliasTheBigMasterofIssues: removed, sorry

EliasTheBigMasterofIssues:
fixed. i wonder...

Issue #289, opened 2022-01-29
controller support


Etaash-mathamsetty:
controller support pls?

Etaash-mathamsetty:
using it over usb works
but 
MY TRIGGERS DON'T WORK :(

electrikjesus:
You likely need to configure the triggers to work using a gamepad keymapper from Android side.

Etaash-mathamsetty:
> You likely need to configure the triggers to work using a gamepad keymapper from Android side.

I just used ds360 (I have a dualsense)
but i'll try that since I don't have any vibration

Etaash-mathamsetty:
ok did some more testing, and I realized that the controller dosen't work unless you plug it in/ connect it AFTER the waydroid session starts, I think that this should probably get fixed
also do you have any good gamepad keymappers that work on waydroid?

DocMAX:
so how can i make my xbox360 controller work?

YogSottot:
> good gamepad keymappers that work on waydroid

Mantis Gamepad Pro 
After loading android you need to execute
```sudo waydroid shell sh /sdcard/Android/data/app.mantispro.gamepad/files/buddyNew.sh```

Also for the gamepad to work you need
```
waydroid prop set persist.waydroid.udev true
waydroid prop set persist.waydroid.uevent true
```

Controller reconnections can be done by a command in the console or by a script, rather than by physically reconnecting the controller. This seems more convenient to me.
```
echo 0 > /sys/bus/usb/devices/3-2/authorized
echo 1 > /sys/bus/usb/devices/3-2/authorized
```
3-2 should be replaced with your device number.

This is for a wired usb controller, of course
To find out the device number, you need to look at vid|pid via ```lsusb```

Then run the command
```
for X in /sys/bus/usb/devices/*; do 
    echo "$X"
    cat "$X/idVendor" 2>/dev/null 
    cat "$X/idProduct" 2>/dev/null
    echo
done

```

Xtr126:
> Controller reconnections can be done by a command in the console or by a script, rather than by physically reconnecting the controller. This seems more convenient to me.
> 
> ```
> echo 0 > /sys/bus/usb/devices/3-2/authorized
> echo 1 > /sys/bus/usb/devices/3-2/authorized
> ```
> 
> 3-2 should be replaced with your device number.
> 
> This is for a wired usb controller, of course To find out the device number, you need to look at vid|pid via `lsusb`
> 
> Then run the command
> 
> ```
[...]

PR #2322, opened 2026-05-20
[WIP] Make controllers detected without needing to replug

Twig6943:
Solves https://github.com/waydroid/waydroid/issues/289

This is still wip as it seems to mess up mouse input for whatever reason. Feedback/help is welcome

Real sorry for the commits by @AhmetCildan he thought he was on a placeholder branch and was testing something git related. That being said we're working on getting both working if possible


tools/actions/container_manager.py
```diff
@@ -100,6 +101,9 @@ def chmod(path, mode):
         perm_list.extend(glob.glob("/dev/video*"))
         # DMA-BUF Heaps
         perm_list.extend(glob.glob("/dev/dma_heap/*"))
+        # Input devices (gamepads, joysticks, etc.)
+        perm_list.extend(glob.glob("/dev/input/event*"))
```
supechicken, line 105: This will pass through all input devices (including mouse/keyboard/touchpad/touchscreen) into Android. Those devices are handled by Wayland/Waydroid and shouldn't exist inside Android

We need a way to check whether the input device is a controller or not before passing it in

Twig6943:
Please see https://github.com/waydroid/waydroid/pull/2345

PR #2369, opened 2026-07-24
initializer: Add alternate version detection logic for Halium 15+

JamiKettunen:
[VNDK is deprecated](https://source.android.com/docs/core/architecture/vndk#vndk-deprecation) from 15 onwards with its files now largely just present as normal vendor partition contents. On [Volla Phone Quintus with UT port rebased to latest Android firmware on kernel v6.6](https://gitlab.com/ubports/porting/reference-device-ports/halium13/volla-phone-quintus/volla-algiz/-/tree/halium-16.0) `ro.vndk.version` is no longer defined while the best alternative for the same job seems to be `ro.vendor.build.version.sdk=35`.

### DIFF
--- a/tools/actions/initializer.py
+++ b/tools/actions/initializer.py
@@ -21,6 +21,7 @@ def is_initialized(args):
 
 def get_vendor_type(args):
     vndk_str = helpers.props.host_get(args, "ro.vndk.version")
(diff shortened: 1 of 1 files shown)


aleasto:
Yep, thanks

Issue #2229, opened 2026-02-17
[Feature] Update image OS to Android 16 / Lineage 23


ItachiSan:
### Is your feature request related to a problem? Please describe.

Recently, no new builds for the OS image has been provided.
The base OS is stuck at Lineage 20 while Lineage 23 is out.

### Describe the solution you'd like.

New OS images with LineageOS 23 are available and used for WayDroid.

### Describe alternatives you've considered.

I've asked why the OS updates have stopped and was mentioned that the builds were provided through another provider (BlissOS I believe?).

I am available to help to restore the builds and/or migrate to e.g. Github CI for the OS builds.

supechicken:
Just got LOS 22.2 running, not for LOS 23 however:

<img width="2879" height="1799" alt="Image" src="https://github.com/user-attachments/assets/3837e24c-f7d8-4c3c-80b4-474b9e14dcc7" />

https://github.com/supechicken/waydroid-builds/releases/tag/20260224

ItachiSan:
@supechicken that is nice!
Is there documentation for building / testing?

ghost:
What were the issues with android 16 @supechicken? There also are Android 17 builds beta so maybe they can be experimented with?

DaanSelen:
I get Waydroid on a clean ubuntu just popping up and then just closing again and again. I cannot get the lineage OS running stably.

Mertsudo:
+1

Sebz102030:
The issiue of waydroid poping and than disapearing mosty i see it when I Tried to force softwear Rendering (i tried on fedora with amd igpu and nvidia dgpu)

Atemo-C:
@DaanSelen & @Sebz102030 These issues are unrelated to the current issue, please see existing opened/closed issues for this.

tajo48:
@DaanSelen @Sebz102030 
I fixed it with this command:
`sudo systemctl stop waydroid-container && sudo rm -rf /var/lib/waydroid ~/.local/share/waydroid ~/.local/share/applications/*waydroid*`

DaanSelen:
> [@DaanSelen](https://github.com/DaanSelen) [@Sebz102030](https://github.com/Sebz102030) I fixed it with this command: `sudo systemctl stop waydroid-container && sudo rm -rf /var/lib/waydroid ~/.local/share/waydroid ~/.local/share/applications/*waydroid*`

I'll try this later!

PR #2386, opened 2026-08-14
ServiceRunner refactor, 1.6.4 fixes, LOS base update, ARM64 translation layer, and release-package workflow

IamAzmathullaShaikh:
## Summary

This PR brings the fork's full delta over upstream `e7d73e7` (7 commits,
~64 files, +4900/−400): a ServiceRunner/dispatch refactor with hardened
root/init guards, the 1.6.4 bug-fix batch, a LineageOS 13→16 support
update, a working ARM64 translation layer with end-to-end verification,
and a release-package workflow with a licensing review.

### ARM64 translation layer (libndk_translation)

- **New `waydroid arm-translation` action** (`install`/`status`/`uninstall`),
  root-gated automatically via the dispatch table.
- **One-command install**: `waydroid arm-translation install` with no flags
  fetches a known-good prebuilt (Android 13, md5-verified), auto-detects
  the `prebuilts/` tree, validates, and installs. `--source`/`--archive`/
  `--url` accept user artifacts.
- **Overlay sync instead of bind mounts**: the container's `/system` overlay
  is mounted read-only, so bind-mounting into it fails silently. Artifacts
  are copied into the overlay lowerdir at session start, before
  `mount_rootfs()` mounts it (the waydroid_script approach).
- **Full native-bridge prop set** in `make_base_props`: emulated
  `ro.product.cpu.abilist*` (fixes both Play's delivery gate and local
  `INSTALL_FAILED_NO_MATCHING_ABIS`), plus `ro.dalvik.vm.isa.arm{,64}` and
  `ro.enable.native.bridge.exec` required by ART and the binfmt_misc
  registration. `[properties]` still overrides.
- **Verified live** (2026-08-14, x86_64 host): an ARM-only APK that failed
  with `-113` before now installs; a test app with a real AArch64 `.so`
  rendered its activity with `ndk_translation: Initialized NDK translation
  (aarch64)` in logcat.

### Release packages + GitHub Actions

- **`.github/workflows/build-release.yaml`**: weekly + manual-dispatch
  workflow that builds a self-contained release package (tooling install
  tree + latest Android images + optional ARM64 translation layer) and
  publishes it as a GitHub Release asset. Inputs select channel
  (`official` / `community-los22` / `community-los23`), arch, system type
  (`VANILLA`/`GAPPS`), and whether to bundle the translation layer.
- **`tools/release/package.py`**: channel resolution, SHA-256 validation
  against the official OTA manifest (checksums computed and recorded for
  community images, which publish none), image extraction, and tarball
  assembly with `install.sh`, `SHA256SUMS`, `LICENSE` and `NOTICE`.
- **`docs/LICENSING.md`**: a licensing/liability review of every bundled
  component — project GPL-3.0-or-later, images LineageOS (Apache-2.0
  framework / GPL-2.0 kernel), and the key finding that the `supremegamers`
  translation prebuilt declares **NOASSERTION** (Google's source is
  Apache-2.0 but the packaging has no explicit grant). Mitigated by
  optional bundling and an explicit NOTICE with provenance + no-warranty
  statement.
- Full LineageOS source builds (~300 GB disk, ~5 h) cannot run on
  GitHub-hosted runners; the workflow packages the upstream-published
  images instead, integrity-verified.

### ServiceRunner + dispatch table

- `tools/services/runner.py` with a dispatch table; `ROOT_ACTIONS` and
  `ALLOWED_UNINITIALIZED` are now **derived from the DISPATCH table's guard
  flags**, so a new action cannot silently skip the root/init guards.
- Session stop ordering, hardening, and expanded test coverage.

### 1.6.4 bug-fix batch

- `@BINDIR@` substitution in the systemd/D-Bus activation files so
  `make install PREFIX=/usr/local` works (was hardcoded `/usr/bin/waydroid`).
- schedtune lazy-unmount retry + actionable warning.
- `enableNFC`/`enableBluetooth` no-ops log at WARNING level.
- `waydroid-net.sh` defaults `LXC_USE_NFT` to auto (nftables-only hosts).
- IPv6 NAT variables now read from the environment.
- Binder/vndbinder/hwbinder config migration in `upgrader.migration()`.
- `waydroid log` path selection extracted and unit-tested.
- Version 1.6.4 + changelog entry; E501 enforced at 131 columns.
- Makefile install prunes `__pycache__` from the install tree.

### LineageOS / Android base support

- README/docs updated: tooling supports Android 13–16 (LineageOS 20–23)
  via runtime API detection; official OTA images remain Android 13 with
  community LineageOS 22/23 builds available (upstream #2229 tracks the
  official rebuild).

## Testing

- `make check`: ruff clean, **128 passed / 1 skipped**.
- `make install DESTDIR=...`: clean tree, `__pycache__` pruned, units
  rendered with correct `ExecStart`.
- Release packaging: all three channels resolve correctly against the live
  endpoints; the staged translation layer matches the runtime `system/`
  layout with exec bits preserved.
- Live runtime verification on a real x86_64 host (binder, Wayland,
  container boot, ARM translation end-to-end).

## Notes for maintainers

- The `--url` default (third-party `supremegamers/...` prebuilt) is
  convenience-only; `--source`/`--archive` remain the primary artifact
  inputs and the code path is identical.
- The `ro.product.cpu.abilist` emulated-ABI advertising is opt-in: it only
  happens when the translation layer is actually installed, and
  `[properties]` in `waydroid.cfg` can override every prop.
- The release workflow publishes to the fork's own Releases page by
  default; publishing to upstream would need maintainer approval.

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -1,0 +1,146 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Build a self-contained Waydroid release package (tooling + latest Android
+# images + optional ARM64 translation layer) and publish it as a GitHub
+# Release asset.
+#
+# The channel input selects the image source:
+#   official          - latest validated build from ota.waydro.id (A13/LOS20)
+#   community-los22   - latest matching WayDroid-ATV build (Android 14)
+#   community-los23   - latest matching WayDroid-ATV build (Android 15/16)
+#
+# Full LineageOS source builds (~300 GB disk, ~5 h) cannot run on
+# GitHub-hosted runners, so this workflow packages the upstream-published
+# images instead, validating their SHA-256 checksums (official channel) or
+# computing and recording them (community channel). See docs/LICENSING.md
+# for the licensing/liability review of everything this packages.
+
+name: Build release package
+
+on:
+  workflow_dispatch:
+    inputs:
+      channel:
+        description: Image source channel
+        type: choice
+        options:
+          - official
+          - community-los22
+          - community-los23
+        default: official
+      arch:
+        description: Target architecture
+        type: choice
+        options:
+          - x86_64
+          - arm64
+        default: x86_64
+      system_type:
+        description: System image variant (GAPPS for Play Services)
+        type: choice
+        options:
+          - VANILLA
+          - GAPPS
+        default: VANILLA
+      include_arm_translation:
+        description: Bundle the ARM64 translation layer (x86_64 hosts)
+        type: boolean
+        default: true
+      version:
+        description: Version string for the package name
+        type: string
+        default: "1.6.4"
+  schedule:
+    # Weekly rebuild of the latest official images.
+    - cron: "30 3 * * 1"
+
+permissions:
+  contents: write
+
+concurrency:
+  group: build-release
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install checks
+        run: |
+          pip install ruff pytest
+
+      - name: Run checks (lint + tests)
+        run: |
+          make check PYTHON=python3
+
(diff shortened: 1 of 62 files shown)


aleasto:
No. If you have valuable changes, please submit them as individual pull requests.

Issue #2266, opened 2026-03-31
[BUG] `waydroid status` is not a reliable method to check status


Samueru-sama:
### Describe the bug

[context](https://github.com/pkgforge-dev/Anylinux-AppImages/issues/421#issuecomment-4152351819)

`waydroid` status can report nothing about the container when it is in fact already running. 

### Waydroid version

1.6.2

### Device

Linux Desktop

### Operating System

any linux

### Kernel version

-

### Desktop Environment

-

### GPU

-

### Logs

-

PR #2391, opened 2026-08-14
release: add GitHub Actions workflow and packaging script for release tarballs

IamAzmathullaShaikh:
Adds a release-package pipeline that assembles a self-contained `.tar.xz` (tooling staged via `make install DESTDIR`, latest Android images, optional ARM64 translation layer) and publishes it to GitHub Releases.

- **channels**: `official` (sha256-validated against the ota.waydro.id manifest) or `community-los22`/`community-los23` (checksums computed and recorded, since upstream publishes none)
- **`tools/release/package.py`**: resolves the channel, validates integrity, stages the tree preserving the `usr/bin/waydroid` symlink, writes a relative-path `SHA256SUMS` and a `NOTICE` with component provenance
- **workflow**: weekly schedule + manual dispatch with channel/arch/system_type/include_arm_translation/version inputs
- **`docs/LICENSING.md`**: license + liability review of every bundled component (GPL-3.0 tooling, Apache-2.0/BSD components, Android images, and the NOASSERTION libndk_translation prebuilt, which is bundled only when explicitly requested)
- README usage instructions and a pytest suite (16 passed)

This is fork-oriented (publishes to the repo's Releases page) but the script and licensing review are portable.

### DIFF
--- a/.github/workflows/build-release.yaml
+++ b/.github/workflows/build-release.yaml
@@ -1,0 +1,146 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Build a self-contained Waydroid release package (tooling + latest Android
+# images + optional ARM64 translation layer) and publish it as a GitHub
+# Release asset.
+#
+# The channel input selects the image source:
+#   official          - latest validated build from ota.waydro.id (A13/LOS20)
+#   community-los22   - latest matching WayDroid-ATV build (Android 14)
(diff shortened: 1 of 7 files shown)


Twig6943:
The app can be packaged as an appimage instead but that depends on [this issue](https://github.com/waydroid/waydroid/issues/2266) , I'm not sure if this PR is worthy of being merged

